PR #506, opened 2016-06-23
some additions that make the automation process go smoother.

snarfmonkey:
couldn't find read-write-group.py in the hub repo, and that globalConfig is helpful.
Let me know if this is not a good place for it.


moulliet:
read-write-group.py was renamed to verifier.py.
I'm not sure that we want to check in FS server names in this repo.

PR #555, opened 2016-07-21
Splitting up the previous/next endpoints

Exide:
This takes the burden off the DataDog request filter and instead uses two explicit endpoints for previous and next.


Exide:
Abandoning this solution in favor of just filtering.

PR #618, opened 2016-09-29
changing error code for protection to 403

moulliet:


Exide:
Looks good. I like the use of ```Response.Status.FORBIDDEN``` instead of the raw code.

PR #650, opened 2016-11-02
One hub to docker them

snarfmonkey:
This is a big one.

* Hub image now based on Debian (FROM: java:8) instead of alpine, mostly for nptq


snarfmonkey:
This is a big one.

* Hub image now based on Debian (FROM: java:8) instead of alpine, mostly for nptq
* Locustwhale image now based on Ubuntu and requires all arguments that the cmd-line does
* This hub can be used for single or clustered hub. Please read the README.
* There is a default logback.xml stuffed in the container image. It just spits logging to stdout. It can be over-ridden with a mounted-in /etc/hub/logback.xml
* Updated the READMEs a bunch. 
* Changed dockerfile to create in-container /mnt/log and /mnt/spoke, just in case
* other stuff

Issue #655, opened 2016-11-03
Fix Docker Rolling Deploy


moulliet:
During a docker rolling deploy, the instances do not give the hub enough time to gracefully shutdown.  This can cause 5xx level errors for clients.

PR #658, opened 2016-11-09
mostly added in 80/8080 into if statement for shutdown curl command.

snarfmonkey:
* add a line into if statement to use port 8080 for curl command(s) in the script
namely to shutdown the hub gracefully (jvm and remove from LB)
* remove "APP" as an input variable because this is just a hub script. simplify.
* change readme and docker-compose and Dockerfile CMD to reflect one less variable
* slightly re-arrange Dockerfile. My logic is to order the steps from least to most-likely to change, thus cache can be used for the earlier steps more often.
Also, since we are running the script with bash, i removed the executable part.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,15 +2,14 @@ FROM java:8
 
 RUN apt-get update && apt-get install -y bash ntp curl
 
-ADD run /opt/hub/bin/run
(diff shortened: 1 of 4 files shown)


moulliet:
fixes #655

PR #742, opened 2017-03-10
Ddog downtime709

### LINKED ISSUE #709: Integrate with DataDog for downtime
During a rolling restart, we know that the hub node might be unavailable for a few minutes.
We can let Datadog know that with this API http://docs.datadoghq.com/api/#downtimes

benFlightStats:
close #709


benFlightStats:
integration test error

PR #788, opened 2017-05-17
Historically large783

benFlightStats:


moulliet:
Since these changes create an item in S3 that is written by S3LargeContentDao and read by S3SingleContentDao, we should make the marshalling shared between the classes.

PR #795, opened 2017-05-24
S3 retry

moulliet:
This will return a 500 response code when the hub encounters networking issues with S3.
It also increases the wait time suggestions to the S3 client.

### DIFF
--- a/src/main/java/com/flightstats/hub/app/HubMain.java
+++ b/src/main/java/com/flightstats/hub/app/HubMain.java
@@ -14,6 +14,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.security.Security;
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


src/main/java/com/flightstats/hub/app/HubMain.java
```diff
@@ -35,6 +36,7 @@ public static void main(String[] args) throws Exception {
     }
 
     static void start() throws Exception {
+        Security.setProperty("networkaddress.cache.ttl", "60");
```
Exide: How about ```ttlSeconds``` or ```ttl.seconds``` for additional context on the number?

Exide:
Looks good :)

src/main/java/com/flightstats/hub/app/HubMain.java
Exide: Nevermind. I thought this was a setting on our side.

PR #848, opened 2017-07-18
Returns item length on GETs

### LINKED ISSUE #771: Provide Content Length headers for Items
This feature should be compliant with https://tools.ietf.org/html/rfc7230#page-30

Exide:
resolves #771 

GET requests now return a length header, ```X-Item-Length```, that equals the raw, uncompressed bytes of the item.

For items inserted prior to this change the value will be ```-1```. If you want to force the system to calculate the value you can provide a header, ```X-Item-Length-Required: true```, with your request.


src/main/java/com/flightstats/hub/dao/aws/S3SingleContentDao.java
```diff
@@ -327,6 +327,7 @@ public static ObjectMetadata createObjectMetadata(Content content, boolean useEn
         } else {
             metadata.addUserMetadata("type", "none");
         }
+        metadata.addUserMetadata("payload", "none");
```
moulliet: What does this do?

Exide: It sets the payload metadata to none obviously :D Honestly I'm not sure what the intent was there. I'm going to double check that its not a loose end that needs tied up and just remove it entirely if so.

Issue #832, opened 2017-07-10
Add NPM plumbing to JS tests


Exide:
Include a ```package.json``` file(s) for dependency management. I want to be able to go into the test directory and type ```npm install``` and have everything needed to run the tests.

Icing on the cake would be to use NPM for test execution. For example ```npm run integration-tests```.

Exide:
We might need to audit the tests and see if they're all still needed.

For example: https://github.com/flightstats/hub/blob/dbe0598/src/test/continuous/hub-s3-costs_spec.js

Exide:
I haven't looked at all of the tests yet but each continuous test takes different parameters. Might make sense to have a ```config.json``` for shared parameters and sane defaults for the others. This would allow us to have a ```npm run alll``` command.

This is sort of what is going on with the integration tests.

Exide:
There is more I want to do with the JS stuff but this issue has been resolved.

PR #883, opened 2017-08-09
Restructured the layout of the JS tests

Exide:
... and set a minimum on Node.js and NPM versions.

**This will break integration, continuous, and long running tests in Jenkins until they're updated.**


Exide:
This helped with #832

PR #887, opened 2017-08-10
Cleanup JS dependencies

Exide:
helps with #832


Exide:
Actually, I want to try one more thing before PR'ing this.

Exide:
Ok, I believe this is ready now.

PR #892, opened 2017-08-15
Reset webhook start item843

benFlightStats:


src/main/java/com/flightstats/hub/webhook/WebhookResource.java
```diff
@@ -168,4 +170,29 @@ static Response deleter(@PathParam("name") String name) {
...
+        return cursorUpdater(name, body, uriInfo);
+    }
+
+    static Response cursorUpdater(@PathParam("name") String name, String body, UriInfo uriInfo) {
+        logger.info("update cursor webhook {} {}", name, body);
+        Webhook webhook = Webhook.fromJson(body, webhookService.get(name)).withName(name);
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            JsonNode root = mapper.readTree(body);
+            if (root.has("item")) {
```
Exide: Instead of passing in a JSON object with the URL...

    {"item": "http://some/item/url"}

.. how about just a plain-text body with the URL?

    http://some/item/url

benFlightStats: @Exide made the change

src/main/java/com/flightstats/hub/webhook/WebhookResource.java
```diff
@@ -168,4 +169,31 @@ static Response deleter(@PathParam("name") String name) {
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.TEXT_PLAIN)
+    public Response updateCursor(@PathParam("name") String name, String body) {
+        return cursorUpdater(name, body, uriInfo);
+    }
+
+    static Response cursorUpdater(@PathParam("name") String name, String body, UriInfo uriInfo) {
+        logger.info("update cursor webhook {} {}", name, body);
+        Webhook webhook = Webhook.fromJson("{}", webhookService.get(name)).withName(name);
+        ObjectMapper mapper = new ObjectMapper();
```
Exide: This isn't being used.

PR #899, opened 2017-08-18
Zero out non active retention policies865

### LINKED ISSUE #865: channel retention policy - zero out non active policies
to make it especially clear which policy is active

benFlightStats:
fixes #865


src/main/java/com/flightstats/hub/model/ChannelConfig.java
```diff
@@ -61,7 +61,10 @@ private ChannelConfig(String name, String owner, Date creationDate, long ttlDays
         this.mutableTime = mutableTime;
         this.allowZeroBytes = allowZeroBytes;
         this.keepForever = keepForever;  // keepForever overrides all other retention policies
-        if (maxItems == 0 && ttlDays == 0 && mutableTime == null) {
+        if (this.keepForever) {
```
moulliet: should this if block also set mutableTime to null?

PR #936, opened 2017-09-26
Webhook tag901

### LINKED ISSUE #901: Allow tag webhooks
In the same way that tags allow query operations on a group of channels, a tag webhook would create webhooks on a group of channels.
As the channel membership of the tag changes, the webhooks would be stopped and started.

This feature would simplify the maintenance of many webhooks all delivering to the same destination.  
Individual webhooks should be accessible for monitoring, pausing and cursor ops.

benFlightStats:
fixes #901


src/main/java/com/flightstats/hub/webhook/WebhookService.java
```diff
@@ -42,6 +42,9 @@ public WebhookService(@Named("Webhook") Dao<Webhook> webhookDao, WebhookValidato
         webhook = webhook.withDefaults();
         webhookValidator.validate(webhook);
         Optional<Webhook> preExisting = get(webhook.getName());
+        if (webhook.isTagPrototype()) {
+            return upsertTagWebhook(webhook, preExisting);
+        }
```
moulliet: Should this be if else?

src/main/java/com/flightstats/hub/webhook/WebhookService.java
```diff
@@ -70,6 +73,13 @@ public WebhookService(@Named("Webhook") Dao<Webhook> webhookDao, WebhookValidato
         return preExisting;
     }
 
+    public Optional<Webhook> upsertTagWebhook(Webhook webhook, Optional<Webhook> preExisting) {
+        webhookDao.upsert(webhook);
+//        webhookManager.notifyWatchers(webhook);
```
moulliet: If we are updating a tag webhook, does that need to be processed by webhookManager?

src/main/java/com/flightstats/hub/webhook/TagWebhook.java
```diff
@@ -0,0 +1,105 @@
...
+    private final static Dao<Webhook> webhookDao = HubProvider.getInstance(
+            new TypeLiteral<Dao<Webhook>>() {
+            }, "Webhook");
+
+    private static Set<Webhook> webhookPrototypesWithTag(String tag) {
+        Set<Webhook> webhookSet = new HashSet<>(webhookDao.getAll(false));
+
+        Predicate<Webhook> withTagName = wh -> tag.equals(wh.getTag());
+        Predicate<Webhook> isTagWebhookPrototype = Webhook::isTagPrototype;
+        return webhookSet.stream().filter(withTagName).filter(isTagWebhookPrototype).collect(Collectors.toSet());
```
moulliet: Can you inline the predicates, and use fluent formatting?

src/main/java/com/flightstats/hub/webhook/Webhook.java
```diff
@@ -294,6 +324,9 @@ public boolean equals(Object o) {
         final Object this$name = this.getName();
         final Object other$name = other.getName();
         if (this$name == null ? other$name != null : !this$name.equals(other$name)) return false;
+        final Object this$tag = this.getTag();
```
moulliet: What do you think about reverting to lombok?

benFlightStats: created an issue to relombok

src/main/java/com/flightstats/hub/webhook/WebhookService.java
benFlightStats: else is not necessary because it is a method exit point

src/main/java/com/flightstats/hub/webhook/Webhook.java
benFlightStats: #938

PR #937, opened 2017-09-26
Spoke store enforcer

Exide:
Does the following on startup unless ```spoke.enforceStore``` is false (defaults to true):

- Verifies ```spoke.path``` exists on the filesystem
- Looks for ```spoke.previousPaths``` in the config (comma delimited)
- Loops over the previous paths and, if they exist, merges them into the ```spoke.path``` location

### DIFF
--- a/src/main/java/com/flightstats/hub/dao/file/FileUtil.java
+++ b/src/main/java/com/flightstats/hub/dao/file/FileUtil.java
@@ -104,7 +104,7 @@ public class FileUtil {
             Path path = paths.get(i);
 
             if (!Files.exists(path)) {
-                logger.info("skipping, already moved {}", path);
+                logger.debug("skipping, already moved {}", path);
                 continue;
             }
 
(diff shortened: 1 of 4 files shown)


Exide:
Running more tests. Will reopen when complete.

Exide:
I ran this against DLS with the following situations:

- Moved store to a child directory of the previous location
- Moved store to a parent directory of the previous location

Both were successful.

src/main/java/com/flightstats/hub/dao/file/FileUtil.java
```diff
@@ -70,9 +72,57 @@ static boolean write(byte[] bytes, String filename, String path) {
...
+        List<Path> paths = Files.walk(source).collect(Collectors.toList());
+        for (int i = 0; i < paths.size(); ++i) {
+            Path path = paths.get(i);
+
+            if (!Files.exists(path)) {
+                logger.debug("skipping, already moved {}", path);
+                continue;
+            }
+
+            if (path.startsWith(destination)) {
```
moulliet: I'm not sure this is correct.

Exide: Agreed. Closing this PR again until I can make sure its bulletproof or not needed at all :p

PR #949, opened 2017-10-19
Balancing

### LINKED ISSUE #935: Even out webhook balancing
The strategy from #701 can be improved.
Currently, the oldest running process will have the most, and the youngest process will have the least.

moulliet:
This balances based on the number of webhooks 
fixes #935

### DIFF
--- a/src/main/java/com/flightstats/hub/webhook/WebhookManager.java
+++ b/src/main/java/com/flightstats/hub/webhook/WebhookManager.java
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.*;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/flightstats/hub/webhook/WebhookManager.java
```diff
@@ -95,41 +95,78 @@ private void manageWebhook(Webhook daoWebhook, boolean webhookChanged) {
...
                 callOneRun(name, servers);
             }
         } else {
             logger.debug("found new v2 webhook {}", name);
-            callOneRun(name, hubCluster.getRandomServers());
+            callOneRun(name, getOrderedServers());
         }
     }
 
+    /**
```
benFlightStats: If the balancing doesn't end up giving the amount of improvement we want, I'm wondering if some other metric that corresponds to webhook load would be a good alternative.

PR #964, opened 2017-11-07
Spoke backed read cache

Exide:
# WARNING: Before merging this PR you must disable the automated deploys in all staging environments!

This stores batch payloads from S3 to speed up sequential GETs on single items within the batch.

This code is running in ```dls.dev```, ```iad.dev```, and ```pdx.dev```.

PR #978, opened 2017-12-01
Remove migration code (phase one)

Exide:
There are multiple deploys needed to remove all of the migration code. This is the first one and it eliminates old ```RemoteSpokeStore``` methods. It also removes temporary logic in the ```S3BatchResource``` that tries the old location parameter, ```CACHE```.


Exide:
This was to clean up old code after #964.

PR #979, opened 2017-12-02
Added some info on inserting zipped data

Exide:


Exide:
I get asked this question periodically and I wanted to be able to point people to the documentation.

PR #980, opened 2017-12-06
Inclusive

### LINKED ISSUE #976: Allow for /next/n to be inclusive
We have a pattern where, given a hub uri, we want to get the next `n` items inclusive of the given hub uri.

Something like:
`/channel/foo/2017/11/17/00/00/00/000/abcdef/next/1000?inclusive=true&bulk=true`

moulliet:
fixes #976


tlehman:
Nice work! I ran into the inclusive problem back in October and just worked around it.

PR #983, opened 2017-12-07
Assume we're in a clustered environment unless proven otherwise.

Exide:
Now the test should work properly even if the Hub under test doesn't have a ```hub.type``` entry in its config.

### DIFF
--- a/src/test/javascript/integration/webhook_localhost_spec.js
+++ b/src/test/javascript/integration/webhook_localhost_spec.js
@@ -27,7 +27,7 @@ describe(__filename, function () {
                 expect(response.statusCode).toEqual(200);
                 let hubType = response.body.properties['hub.type'];
                 if (hubType !== undefined) {
-                    isClustered = response.body.properties['hub.type'] == 'aws';
+                    isClustered = hubType === 'aws';
                 }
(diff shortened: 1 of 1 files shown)


src/test/javascript/integration/webhook_localhost_spec.js
```diff
@@ -18,14 +18,17 @@ var webhookResource = utils.getWebhookUrl() + "/" + webhookName;
...
 
     it('determines if this is a single or clustered hub', (done) => {
         let url = `${hubUrlBase}/internal/properties`;
         utils.httpGet(url)
             .then(response => {
                 expect(response.statusCode).toEqual(200);
-                isClustered = response.body.properties['hub.type'] == 'aws';
+                let hubType = response.body.properties['hub.type'];
+                if (hubType !== undefined) {
+                    isClustered = response.body.properties['hub.type'] == 'aws';
```
moulliet: Should this be === ?

Exide: In this case it doesn't matter since they're both strings. However it should be comparing ```hubType``` and not querying the response.body object again.

Commit incoming :)

PR #1001, opened 2017-12-15
Max attempts

### LINKED ISSUE #986: Feature Request: give up on callback after X tries/Y time passed
Dev Services, for reworking the eventing queue management, would like a callback configure where the hub would give up on a callback if it fails more than X times. A similar variant, give up if a callback is older than Y amount of time.

The concrete uses cases are:
* If a customer fails to successfully handle an event posted to their endpoint 5 times, we throw away that event
* If an event is more than 15 minutes delays, the customer doesn't want it anymore, we can throw it away.

Exide:
resolves #986


Exide:
Flaw in the plan. The error array i'm checking is pruned to only 10 items causing the check to return false negatives.

Exide:
Throwing an exception from the StopStrategy to communicate we've reached maxAttempts.

PR #1008, opened 2018-01-03
Fixes the webhook max attempt integration test

Exide:


Exide:
This makes the test use a callback URL unique to each run of the test, eliminating the possibility of receiving irrelevant messages. It also includes some extra logging for better visibility into the test's execution.

PR #1029, opened 2018-02-05
Allow webhooks to publish errors to a Hub channel

### LINKED ISSUE #785: Add the ability to specify a channel on webhooks that would store items dropped by the webhook
When a channel item is older than the ttl of a webhook that channel item is dropped without any record except as an error shown on the webhook, which does not persists.

It would be beneficial if a webhook could be configured so channel items that are dropped due to the webhook ttl would be placed on another channel for analysis.

Exide:
resolves #785


src/main/java/com/flightstats/hub/webhook/Webhook.java
```diff
@@ -191,7 +197,7 @@ boolean isManagedByTag() {
     static Webhook instanceFromTagPrototype(Webhook whp, ChannelConfig channel) {
         String channenUrl = RequestUtils.getHost(whp.getTagUrl()) + "/channel/" + channel.getName();
         String whName = "TAGWH_" + whp.getTagFromTagUrl() + "_" + channel.getName();
-        return new Webhook(whp.callbackUrl, channenUrl, whp.parallelCalls, whName, null, whp.batch, whp.heartbeat, whp.paused, whp.ttlMinutes, whp.maxWaitMinutes, whp.callbackTimeoutSeconds, whp.fastForwardable, null, whp.getTagFromTagUrl(), whp.maxAttempts);
+        return new Webhook(whp.callbackUrl, channenUrl, whp.parallelCalls, whName, null, whp.batch, whp.heartbeat, whp.paused, whp.ttlMinutes, whp.maxWaitMinutes, whp.callbackTimeoutSeconds, whp.fastForwardable, null, whp.getTagFromTagUrl(), whp.maxAttempts, whp.errorChannelUrl);
```
chriskessel: Don't know if you lombok, but there's a toBuilder=true that gives you a copy constructor that's cool, so you could do ```whp.toBuilder().foo(overridefoovalue).bar(overridebarvalue).build();```

Also, typo channenUrl -> channelUrl.

src/main/java/com/flightstats/hub/webhook/WebhookError.java
```diff
@@ -93,6 +102,58 @@ private String getChildPath(String errorRoot, String child) {
...
+        // todo - workaround for HubHost.getLocalNamePort and HubProperties.getAppUrl not returning usable URLs for dockerized single hub
+        String host = RequestUtils.getHost(attempt.getWebhook().getChannelUrl());
+        return host + "/webhook/" + attempt.getWebhook().getName();
+    }
+
+    private String extractTimestamp(String error) {
+        return error.substring(0, error.indexOf(" "));
+    }
+
+    private String extractMessage(String error) {
```
chriskessel: Looks like the errors have a specific format. Could probably use a comment or a reference to a file with comments on what that format is.

Exide: Good call. Since ```WebhookError.java``` is responsible for creating/parsing those strings I'll add a comment to the file indicating the format.

PR #1044, opened 2018-07-10
Write queue metrics

Exide:


src/main/java/com/flightstats/hub/dao/aws/S3WriteQueue.java
```diff
@@ -59,6 +68,10 @@ private S3WriteQueue() throws InterruptedException {
     private void write() throws InterruptedException {
         try {
             ChannelContentKey key = keys.poll(5, TimeUnit.SECONDS);
+            if (key != null) {
+                metricsService.gauge("s3.writeQueue.used", keys.size());
```
chriskessel: Duplicated string, might deserve a constant?

Exide: I thought about it but I went with consistency for now.

PR #1047, opened 2018-07-13
Node errors

Paul-Hess:
continuation of the last PR, passed dev cycle on iad, there are a couple few lints in here, which generally I've avoided for now, but these were only indentation consistency tweaks. One more to come.

### DIFF
--- a/src/test/javascript/integration/channel_creation_zeroBytes_spec.js
+++ b/src/test/javascript/integration/channel_creation_zeroBytes_spec.js
@@ -1,4 +1,8 @@
 require('../integration_config');
+const {
+    fromObjectPath,
+    getProp,
+} = require('../lib/helpers');
 
 var request = require('request');
 var channelName = utils.randomChannelName();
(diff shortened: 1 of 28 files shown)


src/test/javascript/integration/channel_creation_zeroBytes_spec.js
```diff
@@ -13,52 +17,52 @@ var testName = __filename;
  *
  */
 describe(testName, function () {
-    var returnedBody;
-
+    var returnedBody; // TODO: is this actually being used?
```
Exide: It looks like an over zealous memory optimization from the past :) Since `parse` is being used directly anyway I would just get rid of it.

src/test/javascript/integration/channel_missing_spec.js
```diff
@@ -1,9 +1,10 @@
 require('../integration_config');
+const { getProp } = require('../lib/helpers');
 
 var request = require('request');
-var http = require('http');
+// var http = require('http'); // TODO: not used?
```
Exide: Yeah, looks vestigial. I would remove it.

src/test/javascript/integration/channel_missing_spec.js
```diff
@@ -1,9 +1,10 @@
...
 var channelName = utils.randomChannelName();
-var groupName = utils.randomChannelName();
+// var groupName = utils.randomChannelName();  // TODO: not used?
```
Exide: Same as above, vestigial.

src/test/javascript/integration/channel_protect_requires_owner_spec.js
```diff
@@ -1,6 +1,6 @@
 require('../integration_config');
 
-var request = require('request');
+// var request = require('request'); // TODO: unused?
```
Exide: Looks unused to me. Go ahead and prune it.

Exide:
Looks good. Chop that wood :)

Paul-Hess:
merging this, going to cull the TODO: unused ' s on the next pass

PR #1050, opened 2018-07-19
Node errors

Paul-Hess:
live on IAD

### DIFF
--- a/src/test/javascript/integration/content_type_default_spec.js
+++ b/src/test/javascript/integration/content_type_default_spec.js
@@ -1,4 +1,8 @@
 require('../integration_config');
+const {
+    fromObjectPath,
+    getProp,
+} = require('../lib/helpers');
 
 var channelName = utils.randomChannelName();
 var channelResource = channelUrl + "/" + channelName;
@@ -13,7 +17,7 @@ describe(__filename, function () {
 
         utils.httpPost(url, headers, body)
             .then(function (response) {
-                expect(response.statusCode).toEqual(201);
+                expect(getProp('statusCode', response)).toEqual(201);
             })
             .finally(done);
     });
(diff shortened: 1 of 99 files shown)


src/test/javascript/integration/content_type_default_spec.js
```diff
@@ -27,19 +31,28 @@ describe(__filename, function () {
 
         utils.httpPost(url, headers, body)
             .then(function (response) {
-                expect(response.statusCode).toEqual(201);
-                expect(response.body._links.channel.href).toEqual(channelResource);
-                itemURL = response.body._links.self.href;
+                expect(getProp('statusCode', response)).toEqual(201);
+                const links = fromObjectPath(['body', '_links'], response) || {};
```
Exide: Out of genuine curiosity, why only go down to the `_links` level and then break it into `channel` and `self`?

Couldn't you do something like this?

    const channelHREF = fromObjectPath(['body', '_links', 'channel', 'href'], response);
    expect(channelHREF).toEqual(channelResource);
    
    itemURL = fromObjectPath(['body', '_links', 'self', 'href'], response);

src/test/javascript/integration/content_type_default_spec.js
```diff
@@ -27,19 +31,28 @@ describe(__filename, function () {
...
+                  in the next test could possibly be better handled in a before block
+                  leaving all setup in one place and all assertions in the "it"s
+                */
+                itemURL = self.href;
             })
             .finally(done);
     });
 
     it('verifies the correct Content-Type header is returned', function (done) {
+        if (!itemURL) return done.fail('itemURL failed initialization in previous test');
```
Exide: Good idea. Probably useful in a lot of places :)

src/test/javascript/integration/historical_earliest_item_in_channel_spec.js
```diff
@@ -32,11 +31,13 @@ describe(testName, function () {
...
         utils.postItemQ(historicalItem1)
             .then(function (value) {
-                items.push(value.response.headers.location);
+                const location = fromObjectPath(['response', 'headers', 'location'], value);
+                items.push(location);
                 return utils.postItemQ(historicalItem2);
             })
-            .then(function (value) {
-                items.push(value.response.headers.location);
+            .then(function (nextValue) {
```
Exide: Is there a reason you're calling this argument `nextValue` instead of just `value`? Same with `nextLocation` vs. `location`?

Again, just curious ;)

src/test/javascript/integration/historical_epoch_direction_queries_spec.js
```diff
@@ -45,15 +41,15 @@ describe(testName, function () {
     it('posts historical items to ' + channel, function (done) {
         utils.postItemQ(getFormattedUrl(earliestTime))
             .then(function (value) {
-                items.push(value.response.headers.location)
+                items.push(fromObjectPath(['response', 'headers', 'location'], value))
                 return utils.postItemQ(getFormattedUrl(earliestTime.add(1, 'years')));
             })
-            .then(function (value) {
-                items.push(value.response.headers.location)
+            .then(function (value1) {
```
Exide: Same question as above. Why not use `value` for the arguments in the promise chain?

src/test/javascript/integration/historical_earliest_item_in_channel_spec.js
Paul-Hess: just a preference, some linters will puke on using the same var name within a larger scope even though they are obviously within their own sub-scope. I just think it reads well to name them differently if they are in the same chain.

Exide:
99 changes but a bug _probably_ ain't one :)

Looks good. Let's push it out to all the dev environments and let it run there for a while (24 hours?). With how unstable things have been lets get a trend of successes before we merge.

src/test/javascript/integration/historical_earliest_item_in_channel_spec.js
Exide: Ah, gotcha. Works for me. You own this stuff now so whatever stylistically go with whatever you like :)

src/test/javascript/integration/content_type_default_spec.js
Paul-Hess: I think I just did it this way because it looked better on the page to me at the time, no reason overly compelling

Paul-Hess:
sounds good

PR #1063, opened 2018-08-06
parralel test demo

Paul-Hess:
NOT FOR MERGE
This is just an example I put together for prototyping parallel tests with ava js as the test runner.
Just a conversation piece.


crozierm:
Good idea! Are there tests in the Hub that are (or you foresee being) stateful in a way that they might "clobber" each other? If so, any ideas on how to keep them separate and runnable in any order?

Paul-Hess:
The webhook and webscocket integration tests require more serialization as written.
ava also has a feature to force running certain tests in serial, though test files themselves will still run in parallel.
I think breaking these tests up into feature files, then running all the sets of feature tests together in a kubernetes cluster would do it

PR #1067, opened 2018-08-07
Utils refactor 5

Paul-Hess:

PR #1070, opened 2018-08-08
Utils refactor 6

Paul-Hess:
showing as a compare against 5 to keep the diff smaller, ultimately against master


Paul-Hess:
closing until #1067 is fixed and merged

PR #1081, opened 2018-08-20
Collect age metrics around the S3WriteQueue in a passive way

Exide:


Exide:
@crozierm Can you take a look at this one please?

crozierm:
Let's record two metrics, one for the age of items going into the queue and one for items coming out of the queue.

PR #1085, opened 2018-08-21
Webserver refactor

Paul-Hess:


Exide:
I'm going to hold off reviewing this until the conflicts are resolved.

Paul-Hess:
this and #1086 have passed a day on DLS and are ready for review

src/test/javascript/integration/webhook_error_channel_spec.js
```diff
@@ -109,6 +117,7 @@ describe(__filename, () => {
 
     it('waits for the webhook to give up', async () => {
         const clause = (response) => {
+            console.log('response.body', response.body.errors);
```
Exide: Two things.

a) The string prefix should read `response.body.errors`
b) This line should be after the next line so we can use the `errors` constant that has already been safely extracted

Exide:
Looks good

src/test/javascript/integration/webhook_error_channel_spec.js
Paul-Hess: good point

PR #1086, opened 2018-08-22
Network refactor

Paul-Hess:
This, as part of webserver-refactor #1085 is live on dls dev (finally)


Exide:
I skimmed it since it touched 150 different files :p Looks good.

PR #1089, opened 2018-08-23
Getter refactor

Paul-Hess:


src/test/javascript/package.json
```diff
@@ -1,7 +1,8 @@
 {
   "scripts": {
     "test": "node jasmine",
-    "integration-tests": "npm test -- integration/**/*_spec.js"
+    "integration-tests": "npm test -- integration/**/*_spec.js",
+    "integration-test": "npm test -- integration/**/$FILES"
```
Exide: Is `$FILES` populated automatically by npm based on arguments or something? I've not seen that before.

Exide: Ahh, ok. That makes sense.

PR #1091, opened 2018-08-24
Fix disabled tests

Paul-Hess:
this is probably the first time that one of the PR's has non-insignificant test changes, the three tests here were not running in the build before but have been renamed to not be ignored. (It literally took me this long to realize that the jasmine config file was ignoring anything that was not ending in _spec ...derp)


src/test/javascript/package.json
```diff
@@ -13,14 +12,14 @@
...
     "moment": "^2.22.2",
     "parse-link-header": "^1.0.1",
     "request": "^2.87.0",
     "request-promise-native": "^1.0.5",
     "ws": "^5.2.2"
   },
   "devDependencies": {
-    "express": "^4.16.3"
+    "express": "^4.16.3",
+    "jasmine": "^2.99.0"
```
Paul-Hess: I updated jasmine as the previous version was throwing errors in failed async tests, which resulted in uncaught node exceptions, the update catches the errors internally and fails the test gracefully

src/test/javascript/package.json
```diff
@@ -1,8 +1,7 @@
 {
   "scripts": {
     "test": "node jasmine",
-    "integration-tests": "npm test -- integration/**/*_spec.js",
-    "integration-test": "npm test -- integration/**/$FILES"
+    "integration-tests": "npm test --random=false --fail-fast=false"
```
Paul-Hess: I added some flags here to accommodate the upgrade of jasmine, also due to changes in jasmine.js "npm run test" will run all integration tests by default

src/test/javascript/jasmine.js
```diff
@@ -1,22 +1,15 @@
 const Jasmine = require('jasmine');
 const reporter = require('./lib/ddt_reporter');
```
Paul-Hess: this file just got cleaned up a little, though the only behavior difference is it will now run the integration tests as the default behavior

src/test/javascript/package.json
Paul-Hess: (our pattern of for ex: npm run test -- continuous/other_test_spec.js
will continue to work just like before)

Exide:
Looks good.

PR #1092, opened 2018-08-24
Delete channels

Paul-Hess:
Sorry, last one of this size, promise.
It is the same change in every file though so reviewing one is (almost) reviewing them all
this which includes #1091 is going to DLS presently


Exide:
For 117 changed files this was definitely the easiest to review :) Looks good.

PR #1094, opened 2018-08-29
Locust verifiers

Paul-Hess:


Exide:
Go ahead and merge once it's lived in dev for a while.

PR #1096, opened 2018-09-06
Rolling restart

Paul-Hess:


Paul-Hess:
I am not sure this is in a done state, but opening here for discussion for now.

src/test/javascript/lib/helpers/functional.js
```diff
@@ -58,7 +58,8 @@ const waitForCondition = async (conditionalFunc, timeout = 30000) => {
         const wait = setInterval(() => {
             time += 500;
             const timedOut = time >= timeout;
-            process.stdout.write('… … … … …');
+            process.stdout.write(' … … … … … ');
+            process.stdout.write(' °°°°°°°°° ');
```
Exide: ASCII art 👍

Exide:
Specifically for the rolling restart case we could have data already prepared on a volume that we mount during setup. This would reduce the moving pieces of the test and also speed up the initialization process. We could also re-use this data for other tests. Naively I'm thinking about using an EBS volume  but maybe theres a better solution.

Regardless of that thought, I think we should have a conversation about how to prepare the data for testing.

PR #1097, opened 2018-09-11
Crozierm/null tags

crozierm:
commit 1: auto-format with intellijrains
commit 2: remote unused static field
commit 2: also add 'nulled out tags' to every DawgMetric.


Exide:
Closing this in favor of my branch, [nullOutTags](https://github.com/flightstats/hub/compare/nullOutTags). Neither of which achieve what we want currently so I'll open a PR when its prudent to do so.

PR #1098, opened 2018-09-11
Bugfix ci

Paul-Hess:


src/test/javascript/lib/helpers/random-values.js
```diff
@@ -22,12 +22,15 @@ const randomString = (len) => {
 const randomTag = () => `tag${Math.random().toString().replace(".", "")}`;
 
 const randomItemsFromArrayByPercentage = (arr, percentage) => {
+    const LIMIT = 60 * 1000;
```
Exide: I would make this an argument on the function with a default value of 60k.

Exide:
Looks good

Issue #1028, opened 2018-01-26
Single hub needs a way to identify its external host/port info


Exide:
Given this configuration (defaults and what our CI uses):
```
app.url = http://localhost/
http.bind_port = 80
```

```
rzander@host$ docker ps -a
CONTAINER ID        IMAGE                    COMMAND                  CREATED          STATUS           PORTS                  NAMES
d20dd74beced        flightstats/hub:latest   "/bin/bash /opt/hu..."   1 second ago     Up 1 second      0.0.0.0:8081->80/tcp   hub
```

```
root@container$ hostname
d20dd74beced
```

What we get from the current helper methods is this:

- ```HubHost.getLocalHttpNameUri``` returns ```http://d20dd74beced:80/```
- ```HubHost.getLocalHttpIpUri``` returns ```http://172.17.0.2:80```
- ```HubProperties.getAppUrl``` returns ```http://localhost/```

However what we need is either of these:
```
http://127.0.0.1:8081
```
```
http://localhost:8081
```

PR #1099, opened 2018-09-12
Return the node info as a header

Exide:
REST responses and item's delivered via webhooks now include the following header:

    Hub-Node: <domain>:<port>

### DIFF
--- a/src/main/java/com/flightstats/hub/app/HubResponseFilter.java
+++ b/src/main/java/com/flightstats/hub/app/HubResponseFilter.java
@@ -1,0 +1,16 @@
+package com.flightstats.hub.app;
+
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerResponseContext;
+import javax.ws.rs.container.ContainerResponseFilter;
+import javax.ws.rs.ext.Provider;
+
+@Provider
+public class HubResponseFilter implements ContainerResponseFilter {
+
+    @Override
+    public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) {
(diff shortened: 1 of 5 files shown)


Exide:
This is running in pdx.dev currently.

src/main/java/com/flightstats/hub/app/HubResponseFilter.java
```diff
@@ -0,0 +1,16 @@
...
+import javax.ws.rs.container.ContainerResponseContext;
+import javax.ws.rs.container.ContainerResponseFilter;
+import javax.ws.rs.ext.Provider;
+
+@Provider
+public class HubResponseFilter implements ContainerResponseFilter {
+
+    @Override
+    public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) {
+        responseContext.getHeaders().add("Hub-Node", HubHost.getLocalNamePort());
```
crozierm: Curious - how does this get the real name while inside of docker?

Paul-Hess:
This looks great to me

src/main/java/com/flightstats/hub/app/HubResponseFilter.java
Exide: I believe its a byproduct of Docker's `host` network. If you jump in one of the running Hub containers it thinks it's hostname is that of it's host.

Exide: We're just calling `InetAddress.getLocalHost().getHostName()` in the application code.

Exide: There is an open issue for when running this in the "single" Hub:
https://github.com/flightstats/hub/issues/1028

PR #1109, opened 2018-11-05
verify zip before attempting to write to s3

Paul-Hess:


Paul-Hess:
pushing this to dls now

PR #1110, opened 2018-11-08
Guice for Jersey

Exide:
This takes Guice built dependencies and maps them to HK2 so Jersey can inject them into resources.


Exide:
This has been merged into [batchArchvingEndpoint](https://github.com/flightstats/hub/tree/batchArchivingEndpoint) and is currently running in iad.dev.

crozierm:
This is some relatively deep re-plumbing of the DI and classpath scanning octarinium.  It's going to defy peer review, unless somebody is secretly a Jetty/Jersey/HK2 mastermind (the answer is "no").

The code cleanups are legit! Statictons removed, welcome back lombok, etc... 

But probably best to talk about the testing and deployment plan, to make sure there aren't functional or performance impacts.

src/main/java/com/flightstats/hub/app/HubMain.java
```diff
@@ -3,110 +3,203 @@
...
-        List<Module> modules = new ArrayList<>();
+        HubProvider.setInjector(injector);
+        HubServices.start(HubServices.TYPE.BEFORE_HEALTH_CHECK);
+
+        // build Jetty server
+        Server server = new Server();
+        HttpConfiguration httpConfig = new HttpConfiguration();
+        SslContextFactory sslContextFactory = getSslContextFactory();
+        if (null != sslContextFactory) {
+            httpConfig.addCustomizer(new SecureRequestCustomizer());
```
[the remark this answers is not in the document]
Exide: That's how it was before I got in there and I was trying to limit my changes. It looks like new code because I merged the HubMain and HubJettyServer classes.

Exide:
I agree. I'll get everyone together to chat about it.

This is just the tip of the iceberg. I'll keep pulling from the mega branch of DI changes as I solve prioritized issues.

Exide:
@crozierm @Paul-Hess @lkemmerer I just pushed 3 commits based on our discussion. I'll merge this tomorrow if I don't hear anything :)

crozierm:
🇹🇭 🌡 🎉

crozierm:
Taiwan-ometer-huzzah

src/main/java/com/flightstats/hub/app/GuiceToHK2Adapter.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private final Injector injector;
+
+    GuiceToHK2Adapter(Injector injector) {
+        this.injector = injector;
+    }
+
+    @Override
+    protected void configure() {
+        injector.getBindings().forEach((key, value) -> {
```
lkemmerer: Will this handle singletons properly? I'm not at all familiar with HK2 and only know enough Guice for basic use cases. There's a way to check for singleton annotations in Guice, but I'm not sure how scoping or whatever works in HK2 binding/factories.

Exide: Good question. I'll double check to be sure but I believe as long as the class is annotated with `javax.inject.Singleton` it should be ok. The problem is most of the code uses `com.google.inject.Singleton`. Because this will only effect new code it shouldn't be an issue as long as the dependencies of the new code are using the native injection stuff.

I'll take a look and see if we can check for the Google singleton annotation and bind it as such when setting up HK2. That would at least patch code that hasn't been migrated.

In practice I like to use the native injection stuff except where explicitly dealing with a library (e.g. a Guice module). Part of my larger DI branch involves migrating to the native injection tools.

lkemmerer: I believe guice'll find the native injection annotations, like you said. I think the bindings have some `isSingleton` method on them (or one of their properties) for guice. HK2 wasn't as clear in my brief doc reading. My concern was mostly legacy stuff where there's an annotation that's not being picked up here and we end up with tons of instances that hang around or were supposed to share state (ugh). Sounds like you've got it covered.

lkemmerer: ...late on the follow-up so maybe not useful, but it looks like HK2 supports singleton scope with something like `bindFactory(...).to(..).in(Singleton.class)`

Exide: Sweet! I was hoping I could just chain an `.singleton()` or something to that effect on the end. Thanks for looking that up.

Exide:
I added some unit tests that showed that the existing code handled singletons properly but had a bug with regard to `Named` dependencies, which is now fixed.

src/test/java/com/flightstats/hub/app/GuiceToHK2AdapterTest.java
```diff
@@ -0,0 +1,154 @@
...
+            @Override
+            protected void configure() {}
+
+            @Provides
+            SimpleService providesService() {
+                return new SimpleService();
+            }
+        });
+        ServiceLocator locator = initializeHK2(injector);
+        locator.getService(SimpleService.class);
```
lkemmerer: I'd probably add an assertion against `getName()` or that the service isn't null or the like, just to have an assertion here...

lkemmerer: or possibly that if you initialize two services, you get different instances? I dunno.

Anyway, thanks for humoring my questions about how singleton stuff works. Looks good to me!

Exide: I had convinced myself that relying on `getService` to throw an exception was good enough but after further reading I should be checking for null.

From the JavaDoc:
>@return An instance of the contract or impl.  May return null if there is no provider that provides the given implementation or contract
> @throws MultiException if there was an error during service creation

lkemmerer:
Cool. I'll merge Tuesday. Thank you!

PR #1117, opened 2018-12-12
Add some tests around WebhookManager and ActiveWebhooks

lkemmerer:
I'm not sure this was all strictly necessary, but it helped me understand what was going on in these two classes. I opted to do integration tests for ActiveWebhooks because a lot of what it does is so ZK-specific and I'm not sure its behavior is directly tested anywhere else.

If this looks reasonable, I should probably deploy the branch to a dev cluster and make sure stuff is still behaving as expected. I'm still having problems remembering how the various hub configs change the implementation (local, test, aws...clustered vs single...) and sometimes change which tests run or what's asserted?


src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
```diff
@@ -0,0 +1,110 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import java.util.*;
+
+import static java.util.stream.Collectors.toList;
+
+public class InternalWebhookClient {
```
Exide: I like the decoupling of client and manager.

src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
```diff
@@ -0,0 +1,110 @@
...
+import com.flightstats.hub.util.HubUtils;
+import com.google.common.annotations.VisibleForTesting;
+import com.sun.jersey.api.client.Client;
+import com.sun.jersey.api.client.ClientResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import java.util.*;
```
Exide: We don't have it written down anywhere but we're trying to migrate away from wildcard imports.

src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    Optional<String> runOnServerWithFewestWebhooks(String name) {
+        return runOnOneServer(name, getOrderedServers());
+    }
+
+    Optional<String> runOnOneServer(String name, Collection<String> servers) {
+        return servers.stream()
+                .filter(server -> put(server + "/internal/webhook/run/" + name))
+                .findFirst();
```
Exide: Doesn't this send a `PUT` to every server instead of just one?

src/test/java/com/flightstats/hub/webhook/InternalWebhookClientTest.java
```diff
@@ -0,0 +1,228 @@
...
+import com.sun.jersey.api.client.WebResource;
+import org.junit.Test;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static java.lang.String.format;
+import static org.junit.Assert.*;
```
Exide: Same comment as above regarding wildcard imports.

src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
lkemmerer: `findFirst` short-circuits on the first element that meets the requirement. I added a test to make sure that's true, though if it's not intuitive, I'm happy to switch back to the loop + break.

src/test/java/com/flightstats/hub/webhook/WebhookManagerTest.java
```diff
@@ -0,0 +1,153 @@
...
+import com.sun.jersey.api.client.WebResource;
+import org.junit.Test;
+
+import java.util.Collection;
+import java.util.Optional;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static com.google.common.collect.Sets.newHashSet;
+import static java.lang.String.format;
+import static org.mockito.Mockito.*;
```
Exide: Same comment as above regarding wildcard imports.

src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
lkemmerer: OK. I'll figure out where I can turn that off in IntelliJ. Pretty sure it's a setting in there somewhere.

Exide:
Aside from my comments I like what I see. Nice work.

src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
Exide: Interesting, I didn't know that. Cool. I looked at the unit test for it and it looks good.

src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
Exide: Yeah. It was less intuitive to change than I had hoped but i don't remember the details :)

Exide:
👍

lkemmerer:
i'm holding off on merging this until after the holidays. It's unlikely that there will be any critical bugs to fix in prod over the holidays, but I don't want this hanging around in master in case something comes up.

PR #1118, opened 2019-01-04
Dropwizard metrics

Paul-Hess:


build.gradle
```diff
@@ -109,13 +109,17 @@ dependencies {
...
+    compile 'io.dropwizard.metrics:metrics-jvm:4.0.0'
+    compile 'com.github.davidb:metrics-influxdb:0.9.3'
 
     compile 'com.atlassian.commonmark:commonmark:0.10.0'
 
     testCompile 'com.sparkjava:spark-core:2.7.1'
     testCompile 'junit:junit:4.+'
     testCompile "org.mockito:mockito-all:1.+"
     testCompile 'org.testng:testng:6.+'
+    testCompile group: 'org.hamcrest', name: 'hamcrest-core', version: '1.3'
```
lkemmerer: Aw, I have been missing this library. Thanks for adding it!

docker/aws_hub.properties
```diff
@@ -55,8 +55,8 @@ app.runNtpMonitor=false
 #This is the time of the oldest live item in a hub.  With the exception of historical, we know not to query before
 # this time.
 #app.birthDay=2015/01/01
-# Enable DataDog metrics
-data_dog.enable=false
+# Enable metrics services
+metrics.enable=false
```
lkemmerer: I'm unfamiliar with how we handle all our config management. Are we going to need to loop in Joe to get these changes pushed out or do they get copied as part of our build + deploy pipeline already?

src/main/java/com/flightstats/hub/app/SingleHubBindings.java
```diff
@@ -27,6 +28,7 @@ protected void configure() {
         bind(ContentService.class).to(SingleContentService.class).asEagerSingleton();
         bind(DocumentationDao.class).to(FileDocumentationDao.class).asEagerSingleton();
         bind(ChannelTtlEnforcer.class).asEagerSingleton();
+        bind(JVMMetricsService.class).asEagerSingleton();
```
lkemmerer: Does this work if you move it from here (and the clustered hub bindings) and put it in plain old `HubBindings`?

src/main/java/com/flightstats/hub/metrics/JVMMetricsService.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Inject
+    JVMMetricsService(MetricRegistry registry, HubVersion hubVersion) {
+        this.hubVersion = hubVersion;
+        this.registry = registry;
+        if (HubProperties.getProperty("metrics.enable", false)) {
+            String env = HubProperties.getProperty("app.environment", "dev");
+            String clusterName = HubProperties.getProperty("cluster.tagName", "single.local");
+            // custom configurable reporter
+            influxReporter = InfluxdbReporter.forRegistry(registry)
```
lkemmerer: I think you could move the `InfluxdbReporter` creation out into a provider in `HubBindings`. Then you can inject it, which makes mocking in tests a bit easier....that might not be necessary in this class, but it's a good practice to do as little in constructors as possible.

src/main/java/com/flightstats/hub/metrics/JVMMetricsService.java
```diff
@@ -0,0 +1,118 @@
...
+    private final Logger logger = LoggerFactory.getLogger(JVMMetricsService.class);
+
+    @Inject
+    private MetricRegistry registry;
+
+    @Inject
+    private HubVersion hubVersion;
+
+    private ScheduledReporter influxReporter;
+    private final String protocolScheme = HubProperties.getProperty("influx.protocolScheme", "http");
```
lkemmerer: I'd probably need to fiddle with it a little bit, but my gut instinct is that it might make sense to pull all the configuration for this out into another class and (again) inject it. That'd clear up responsibilities a little bit. Any time I use `@VisibleForTesting` I tend to wonder if my class is doing too much.

src/test/java/com/flightstats/hub/metrics/JVMMetricsTest.java
```diff
@@ -0,0 +1,64 @@
...
+        MetricRegistry registry = new MetricRegistry();
+        metricsService = new JVMMetricsService(registry, hubVersion);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Test
+    public void registeredMetricNamesTest() {
+
+        SortedSet<String> metricNames = metricsService.getRegisteredMetricNames();
+        for (String metricName : metricNames) {
```
lkemmerer: When I read the assertions for this test, it looks we don't care how many metrics names we find and that it'll happily pass if we get an empty set back. Is that the behavior we're expecting?

docker/aws_hub.properties
Paul-Hess: We will need to loop Joe in I believe, these are just for example

src/main/java/com/flightstats/hub/app/SingleHubBindings.java
Paul-Hess: I am sure it will, for some reason I thought that was an interface the other two implemented and mostly ignored it

src/test/java/com/flightstats/hub/metrics/JVMMetricsTest.java
```diff
@@ -0,0 +1,64 @@
...
+            //noinspection unchecked
+            assertThat(metricName, anyOf(containsString("gc"), containsString("memory"), containsString("thread")));
+            assertThat(metricName, containsString("jvm_"));
+        }
+    }
+
+    @Test
+    public void getProtocolTest() {
+        assertThat(metricsService.getProtocol(), is(InfluxdbProtocol.class));
```
lkemmerer: I'd feel more comfortable if we had more control over the input that this method uses so that we can test both sides of the conditional (http vs udp). I think right now we're depending on the hub properties/config file not to change, which can be brittle. If we use the injected configuration class I mentioned above, that'll be easier to accomplish.

src/main/java/com/flightstats/hub/metrics/JVMMetricsService.java
Paul-Hess: sounds good, I'll give that a shot

src/main/java/com/flightstats/hub/metrics/JVMMetricsService.java
Paul-Hess: makes sense

src/test/java/com/flightstats/hub/metrics/JVMMetricsTest.java
Paul-Hess: heh, good point, I'll make this more specific

lkemmerer:
This is really easy to understand and looks good, especially considering you're not familiar with dependency injection nuances. Let me know if any of the comments don't make sense or the changes seem awkward to implement.

src/main/java/com/flightstats/hub/app/SingleHubBindings.java
lkemmerer: Oh, maybe it is. I'll take a look.

docker/aws_hub.properties
Exide: If you have write permissions (which you should) you can update the configs for each environment within the salt-states BitBucket repo:

https://bitbucket.flightglobal.io/projects/PDX-SRE/repos/salt-states

src/main/java/com/flightstats/hub/app/SingleHubBindings.java
Exide: Yup. `HubBindings` is intended to be a place that all configurations of the Hub use. I would put this there as well.

src/main/java/com/flightstats/hub/metrics/JVMMetricsService.java
Exide: Agreed. One of the paradigms I was trying to move towards was having a POJO that represented configuration and then inject it to those that needed it.

We should probably mark the `HubProperties` class as deprecated as a reminder that its not going to live through the future.

lkemmerer: I almost went on a tangent about how I wish we'd get rid of `HubProperties` and thought better of it. 😃

Exide:
All in all I like what I see. I agree with Laurie's comments. Good work.

src/test/javascript/integration/jvm_metrics_service_spec.js
```diff
@@ -0,0 +1,119 @@
...
+    'cluster=local-single',
+    'env=single',
+    'role=hub',
+    'team=ddt',
+    'version=local',
+];
+
+describe(__filename, function () {
+    beforeAll(async () => {
+        pending('infrastructure needed to allow metrics.enable to be dynamically true in test envs');
```
Paul-Hess: NOTE: this is requires some separate infra changes that I think are out of the scope of this PR, and may be better as one of the tests that is run against an ephemeral cluster constructed for the single test. I see value in the test being written and passing locally in the mean time. Leaving it pending for now...

src/main/java/com/flightstats/hub/metrics/JVMMetricsService.java
```diff
@@ -0,0 +1,57 @@
...
+    public void startUp() {
+        if (metricsConfig.enabled()) {
+            logger.info(
+                    "starting jvm metrics service reporting to {} :// {} : {}",
+                    metricsConfig.getInfluxdbProtocol(),
+                    metricsConfig.getInfluxdbHost(),
+                    metricsConfig.getInfluxdbPort());
+            int intervalSeconds = metricsConfig.getReportingIntervalSeconds();
+            String metricPrefix = "jvm_";
+            metricsRegistry.register(metricPrefix + "gc", new GarbageCollectorMetricSet());
```
crozierm: TBH I didn't think you had to do this by hand.

src/main/java/com/flightstats/hub/metrics/JVMMetricsService.java
```diff
@@ -0,0 +1,57 @@
...
+import com.codahale.metrics.ScheduledReporter;
+import com.google.common.util.concurrent.AbstractIdleService;
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import org.slf4j.LoggerFactory;
+import org.slf4j.Logger;
+import java.util.concurrent.TimeUnit;
+
+@Singleton
+public class JVMMetricsService extends AbstractIdleService {
```
crozierm: IMHO a better name for this would be something like "MetricsLifecycle", as it's not really "about" JVM metrics, it's more about setting up the Metrics Registry, the publisher, and the registering the shutdown hooks

src/main/java/com/flightstats/hub/metrics/MetricsConfig.java
```diff
@@ -0,0 +1,145 @@
...
+    private Properties properties;
+
+    @Getter
+    private final String role = "hub";
+
+    @Getter
+    private final String team = "ddt";
+
+    @Getter(lazy = true)
+    private final String env = safeGetProperty("app.environment", "dev");
```
crozierm: I would probably use a builder pattern for this instead of deferring until invocation with lazy.
Cuz:
A. failing early is better
B. logic in a a getter/setter is considered an anti-pattern by some (I'm like 70/30 on this)

src/main/java/com/flightstats/hub/metrics/MetricsConfig.java
```diff
@@ -0,0 +1,145 @@
...
+public class MetricsConfig {
+    private final Logger logger = LoggerFactory.getLogger(MetricsConfig.class);
+
+    private Properties properties;
+
+    @Getter
+    private final String role = "hub";
+
+    @Getter
+    private final String team = "ddt";
```
Paul-Hess: I am going to make this configurable and remove the ref to internal team name

src/main/java/com/flightstats/hub/app/HubBindings.java
```diff
@@ -220,9 +258,11 @@ protected void configure() {
         bind(FinalCheck.class).to(SpokeFinalCheck.class).asEagerSingleton();
         bind(InFlightService.class).asEagerSingleton();
         bind(ChannelService.class).asEagerSingleton();
-        bind(MetricsConfig.class).toInstance(metricsConfig);
-        bind(MetricRegistry.class).toInstance(metricRegistry);
-        bind(JVMMetricsService.class).asEagerSingleton();
+        bind(MetricsConfig.class)
```
lkemmerer: this maaaay auto-bind as needed, but I'm not sure about that.

Paul-Hess: If I get your drift, it should. I added this in order to use the annotation with the assumption that at a later date we could reuse this metrics config for our other metrics reporters. Would it be cleaner to make that change when we get to that point?

src/main/java/com/flightstats/hub/metrics/MetricsConfig.java
```diff
@@ -22,4 +27,38 @@
     int reportingIntervalSeconds;
     String role;
     String team;
+
+    private static String parseAppVersion() {
```
lkemmerer: I need to think about a better way to do this;... we can possibly inject `HubVersion` into the `MetricsConfigBuilder` and then it could be mocked? Hmm. I might also move this into the builder class if that's the only place it's being used.

src/test/java/com/flightstats/hub/metrics/MetricsConfigTest.java
```diff
@@ -12,161 +15,26 @@
...
-        assertEquals(metricsConfig.getInfluxdbUser(), "");
-        assertEquals(metricsConfig.getInfluxdbDatabaseName(), "hubmain");
-        assertEquals(metricsConfig.getClusterTag(), "local-dev");
-        assertEquals(metricsConfig.getReportingIntervalSeconds(),  15);
+
+        String clusterTag = HubProperties.getProperty("cluster.location", "local") +
+                "-" +
+                HubProperties.getProperty("app.environment", "dev");
+
+        assertEquals(HubProperties.getProperty("metrics.tag.role", "hub"), metricsConfig.getRole());
```
lkemmerer: Tests like this always bug me just a little bit. It'll go away once we clean up the config design, so I'm cool with it for now.

src/main/java/com/flightstats/hub/app/HubMain.java
```diff
@@ -139,6 +144,17 @@ public Server startServer() throws Exception {
         return server;
     }
 
+    List<AbstractIdleService> getBeforeHealthCheckServices(Injector injector) {
+        List<AbstractIdleService> beforeHealthCheckServices = new ArrayList<>();
+        InfluxdbReporterLifecycle influxdbReporterLifecycle = injector.getInstance(InfluxdbReporterLifecycle.class);
+        beforeHealthCheckServices.add(influxdbReporterLifecycle);
```
lkemmerer: How do you feel about instantiating + returning the list in one statement? Java 9 has `List.of(elements...)`, but for now we can use com.google.common's `return newArrayList(servicesAsArgumentList...)`.

chriskessel: FWIW, in pre-9 I use Arrays.asList(...);

lkemmerer: Oh, sweet! Somehow I missed out on this for like a year and a half.

Paul-Hess: Oh cool, TIL thanks!

src/main/java/com/flightstats/hub/metrics/MetricsConfig.java
Paul-Hess: I'll work on cleaning this up

lkemmerer: I think I'm using a provider for the s3 verifier config I extracted out, so that'd be neat for consistency's sake.

PR #1120, opened 2019-01-10
Uniformly remove all zk cruft when we delete a webhook

lkemmerer:
We delete webhooks in a couple places and have some ZK state that we don't currently clean up consistently, so this fixes that. This doesn't affect prod much, since we don't delete webhooks that often, but dev had a pretty massive number of orphaned webhook state.

Other than the first commit bug fix, there's some pretty serious refactoring to `WebhookError`, which might not have been strictly necessary, but the single test it had bugged me and I wanted to use the ZK utils class there, too.

This is currently based on the PR y'all just reviewed. When that's merged, I'll change it to be based on master. Planning to merge that one after we deploy tomorrow.

### DIFF
--- a/src/main/java/com/flightstats/hub/webhook/WebhookError.java
+++ b/src/main/java/com/flightstats/hub/webhook/WebhookError.java
@@ -7,21 +7,22 @@ import com.flightstats.hub.model.Content;
 import com.flightstats.hub.util.RequestUtils;
 import com.flightstats.hub.util.StringUtils;
 import com.flightstats.hub.util.TimeUtil;
+import com.flightstats.hub.util.SafeZooKeeperUtils;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
-import org.apache.curator.framework.CuratorFramework;
-import org.apache.zookeeper.KeeperException;
-import org.apache.zookeeper.data.Stat;
 import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
+import java.time.Duration;
 import java.util.List;
-import java.util.SortedMap;
-import java.util.TreeMap;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
 
 import static com.flightstats.hub.util.RequestUtils.getChannelName;
+import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.toList;
 
 /**
  * This class is responsible for creation and management of webhook error strings.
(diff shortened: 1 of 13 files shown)


src/main/java/com/flightstats/hub/webhook/error/WebhookErrorService.java
```diff
@@ -0,0 +1,74 @@
...
+                                .creationTime(new DateTime(dataWithStat.getStat().getCtime()))
+                                .data(dataWithStat.getData())
+                                .build()))
+                .flatMap(maybeData -> maybeData.map(Stream::of).orElse(Stream.empty()))
+                .collect(toList());
+
+    }
+
+    @VisibleForTesting
+    public static class ErrorNodeNameGenerator {
```
lkemmerer: Bleh, I put this here so I could use deterministic error ids in the test, but it makes me sad.

src/main/java/com/flightstats/hub/webhook/WebhookError.java
```diff
@@ -37,80 +31,42 @@
...
 
-    private String getChildPath(String errorRoot, String child) {
-        return errorRoot + "/" + child;
-    }
+    private List<com.flightstats.hub.webhook.error.WebhookError> limitChildren(String webhook) {
+        List<com.flightstats.hub.webhook.error.WebhookError> errors = webhookErrorService.getErrors(webhook);
 
-    public List<String> get(String webhook) {
-        return limitChildren(webhook);
+        List<com.flightstats.hub.webhook.error.WebhookError> prunedErrors = webhookErrorPruner.pruneErrors(webhook, errors);
```
lkemmerer: I dislike having two `WebhookError` classes (one for data in the `webhook.error` package and then this one). If anyone has name change suggestions, that would be great.

lkemmerer: OK, I renamed some classes. It's better, but not good. 🤷‍♂️

src/main/java/com/flightstats/hub/webhook/WebhookErrorService.java
```diff
@@ -0,0 +1,124 @@
...
+        webhookErrorStateService.add(webhook, error);
+        limitChildren(webhook);
+    }
+
+    public void delete(String webhook) {
+        logger.info("deleting webhook errors for " + webhook);
+        webhookErrorStateService.deleteWebhook(webhook);
+    }
+
+    public List<String> get(String webhook) {
```
crozierm: lookup?

src/main/java/com/flightstats/hub/webhook/WebhookErrorService.java
```diff
@@ -0,0 +1,124 @@
...
+        return limitChildren(webhook).stream()
+                .map(WebhookError::getData)
+                .collect(toList());
+    }
+
+    private List<WebhookError> limitChildren(String webhook) {
```
crozierm: trimAndLookup?

src/main/java/com/flightstats/hub/webhook/error/WebhookErrorStateService.java
```diff
@@ -0,0 +1,74 @@
...
+
+import javax.inject.Inject;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Stream;
+
+import static java.util.stream.Collectors.toList;
+
+public class WebhookErrorStateService {
```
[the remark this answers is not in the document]
lkemmerer: I thought about using DAO, but we were using that for a different backing storage (S3/file or dynamo? I don't remember) and I didn't want to be additionally confusing. Not that this isn't additionally confusing. Repository might work better...if that word doesn't have existing meaning that's different from what we intend here. I'll take a look.

src/main/java/com/flightstats/hub/webhook/WebhookError.java
```diff
@@ -37,80 +38,40 @@
...
+    private final SafeZooKeeperUtils zooKeeperUtils;
     private final ChannelService channelService;
+    private final ErrorNodeNameGenerator errorNameGenerator;
+    private final WebhookErrorReaper webhookErrorReaper;
 
     @Inject
-    public WebhookError(CuratorFramework curator, ChannelService channelService) {
-        this.curator = curator;
-        this.channelService = channelService;
+    public WebhookError(SafeZooKeeperUtils zooKeeperUtils, ChannelService channelService) {
```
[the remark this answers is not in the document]
lkemmerer: I think I do that in a later commit? Well, `WebhookErrorReaper`, anyway. The name generator stuff felt weird because I had really only extracted it into its own class so I could mock it for testing. Pretty sure normal people would have left it as a private method or something. I like the idea of having a single constructor, so I'll make it so.

src/main/java/com/flightstats/hub/webhook/WebhookError.java
```diff
@@ -133,66 +135,12 @@ String generateName() {
         }
     }
 
+    @Builder
+    @Getter
     private static class Error {
         String name;
```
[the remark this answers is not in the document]
lkemmerer: ...which I do in a later commit.

src/main/java/com/flightstats/hub/webhook/WebhookErrorService.java
crozierm: The eternal refactoring struggle... is it too far or not far 'nuff? ;-)

src/main/java/com/flightstats/hub/webhook/error/WebhookError.java
```diff
@@ -0,0 +1,13 @@
+package com.flightstats.hub.webhook.error;
+
+import lombok.Builder;
+import lombok.Value;
+import org.joda.time.DateTime;
+
+@Builder
+@Value
+public class WebhookError {
+    String name;
```
crozierm: 👍 Value does make it final and private

src/main/java/com/flightstats/hub/webhook/WebhookError.java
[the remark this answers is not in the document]
lkemmerer: Oh, yeah. That's why I normally use `@Value`. I hadn't remembered why I did that. Thanks!

src/test/javascript/integration/webhook_max_attempts_spec.js
```diff
@@ -108,6 +108,7 @@ describe(__filename, () => {
         const response = await hubClientGet(webhookResource, headers);
         expect(getProp('statusCode', response)).toEqual(200);
         const body = getProp('body', response) || {};
+      console.info(body);
```
Paul-Hess: sorry for the nitpick, I think we can indent/or remove this line.

lkemmerer: Whoops! Not a nit-pick at all. I'll get rid of it.

lkemmerer:
@crozierm I think I addressed all of your concerns.

PR #1122, opened 2019-01-15
Re-structured how configs are stored and utilized with Docker.

Exide:
Just some QoL tweaks to configs and how we build local Docker images.


configs/read-only-hub.properties
```diff
@@ -0,0 +1,35 @@
...
+
+# spoke
+spoke.write.path=/mnt/spoke/write
+spoke.read.path=/mnt/spoke/read
+spoke.write.ttlMinutes=60
+spoke.read.ttlMinutes=10
+
+# aws
+aws.signing_region=us-east-1
+aws.credentials=/etc/hub/credentials.properties
```
lkemmerer: Dumb question, but does this need to get updated to `read-only-credentials.properties`? 

Also, I'm guessing the AWS credentials files aren't here because they're being managed by salt right now and docker stuff doesn't run in that mode... if we run locally in AWS mode, we'd need to put one somewhere manually, too?

lkemmerer:
lgtm. I'm extrapolating that we'll continue moving non-docker configs/examples here as we clean up our configuration stuff.

configs/read-only-hub.properties
Exide: a) It's a little confusing but the original file name, `read-only-credentials.properties`, gets renamed to `credentials.properties` when it's put into the Docker image.

b) You are correct. I can see the read-only credentials being in salt also but since I was still hacking around I had them in the repo (with fake info) that I have a modified version of locally that actually works.

Exide:
That is my thought as well. This only extract the Docker specific stuff but the rest should go here also.

PR #1124, opened 2019-01-17
increase s3 request timeout, emit metrics/logs on timeout

Paul-Hess:


Paul-Hess:
That's an excellent question, I was supposed to change it to be configurable in pairing with Crozier, and we started down a rabbit hole from there, from which we decided to take a step back until after a longer consultation with you on scope, I'll add that one change in before merging

lkemmerer:
Looks good, thanks for making it configurable!

PR #1126, opened 2019-01-29
Fix docs-related security vulnerabilities

lkemmerer:
Just updated github-pages. I think everything else should work just fine...


lkemmerer:
This is minor, so since Paul looked at it, I'll merge and deal with any doc-building-related fallout.

PR #1127, opened 2019-01-30
Dropwizard metrics

Paul-Hess:
I thought I could just re-open this, but because I force pushed it force pushed me to open it as a new one, here is the link to the previous comments

https://github.com/flightstats/hub/pull/1118


src/test/java/com/flightstats/hub/metrics/MetricsConfigTest.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertFalse;
+
+
+
+public class MetricsConfigTest {
+    @Test
+    public void testDefaultValues_defaultTags() {
```
Paul-Hess: With the other changes I ended up making, it made sense to change this to testing for static expectations after all.

src/main/java/com/flightstats/hub/app/HubBindings.java
```diff
@@ -177,6 +212,12 @@ protected void configure() {
         bind(InFlightService.class).asEagerSingleton();
         bind(ChannelService.class).asEagerSingleton();
 
+
+        bind(HubVersion.class).toInstance(new HubVersion());
```
Paul-Hess: I can't find any reason this wasn't bound in this manner before

lkemmerer: My understanding of guice was that it auto-wired dependencies if there wasn't any fancy settings required (like  special scoping or having to specify a provider or an implementation of an interface)... Was this throwing a DI error without the bind?

src/main/java/com/flightstats/hub/app/HubBindings.java
```diff
@@ -151,6 +160,32 @@ public static ObjectMapper objectMapper() {
         return mapper;
     }
 
+    @Provides
+    @Singleton
+    @Inject
+    public static MetricsConfig metricsReportersConfig(HubVersion hubVersion) {
```
lkemmerer: One final tweak, I think:
During our talk about naming things, you and Crozier brought up using `*Provider`s. I was really only used to putting providers in modules like this, but I like how you have the `InfluxdbReporterProvider`...Having `*ConfigProvider`s would let us test things with logic in them (like `clusterTag` here). Would you be up for moving this into a class that implements `Provider`?

src/test/java/com/flightstats/hub/metrics/MetricsConfigTest.java
```diff
@@ -0,0 +1,54 @@
+package com.flightstats.hub.metrics;
```
lkemmerer: ...so that was not my last comment, I guess... sorry!
This kind of looks like it's solely testing Lombok at this point. It might be worthwhile to borrow some code here for tests around the `InfluxdbReporterProvider`.

lkemmerer:
You've been really patient while we all work through naming and sorted out how we want to do DI for our new config patterns. I feel like this branch has kind of been the guinea pig for that. I think if you'd like to merge as-is, that's fine, or if you want to do /one-more/ round of changes, that's good, too.

src/main/java/com/flightstats/hub/app/HubBindings.java
Paul-Hess: It will error now if not bound, as there is no reference to the injector available here at this point in the lifecycle . Before however I think it would have been easier to bind it for testing ease, as well as the fact we are calling injector.getInstance a couple times (through HubProvider) to get it. I don't know the rule of thumb on that but my first thought was that if calling more than once it might be cleaner to bind it. How do you feel about that?

src/main/java/com/flightstats/hub/app/HubBindings.java
Paul-Hess: Absolutely, easy enough, and def testing advantage

src/test/java/com/flightstats/hub/metrics/MetricsConfigTest.java
Paul-Hess: No problem, you're right

src/main/java/com/flightstats/hub/app/HubBindings.java
lkemmerer: That works for me. I think I was having issues wrapping my head around the problem when you explained it before...maybe because of all the statics we have with HubProvider and HubProperties? Thanks.

Paul-Hess:
the commits after https://github.com/flightstats/hub/pull/1127#commits-pushed-0b93fff are new and could use fresh review. They are easy to view commit by commit.

3a3f351 is the one that makes the most impact

src/main/java/com/flightstats/hub/metrics/StatsDFilter.java
```diff
@@ -50,15 +50,16 @@ public boolean isTestChannel(String channel) {
     }
 
     boolean isSecondaryReporting(String name) {
-        Optional<Webhook> optionalWebhook = Optional
-                .ofNullable(webhookConfigDao.getCached(name))
-                .filter(Webhook::isSecondaryMetricsReporting);
+        if (name.equals("request")) return false;
 
-        Optional<ChannelConfig> optionalChannelConfig = Optional
+        return Optional
```
lkemmerer: I think I would have left this as two variables instead of inlining them... I know the fact that you can't duck-type these is annoying. Another option might be to do something like...  (note: almost certainly has typos:)
```
Optional<Boolean> webhookHasSecondaryReporting = Optional.ofNullable(webhookConfigDao.getCached(name))
    .map(Webhook::isSecondaryMetricsReporting)
    .orElse(false);
boolean channelHasSecondaryReporting = Optional.ofNullable(channelConfigDao.getCached(name))
    .map(ChannelConfig::isSecondaryMetricsReporting)
    .orElse(false);
return webhookHasSecondaryReporting || channelHasSecondaryReporting;
```
Assuming the DAOs don't have to follow a generic interface, I would also have probably extracted the two variables out as methods into their respective DAOs so they could be tested slightly more directly.

src/main/java/com/flightstats/hub/metrics/StatsdReporter.java
```diff
@@ -36,23 +36,23 @@ public void insert(String channel, long start, ChannelType type, int items, long
 
     public void event(String title, String text, String[] tags) {
         Event event = statsDFormatter.buildCustomEvent(title, text);
-        reportWithFilteredClients(title, (statsDClient -> statsDClient.recordEvent(event, tags)));
+        reportWithFilteredClients(title, (statsDClient) -> statsDClient.recordEvent(event, tags));
```
lkemmerer: Stylistically, I'd usually drop the parentheses around `statsDClient` since there's only one argument so they're not necessary.

src/main/java/com/flightstats/hub/metrics/StatsDFilter.java
```diff
@@ -62,6 +65,21 @@ boolean isSecondaryReporting(String name) {
...
     }
 
+    String extractName(String[] tags) {
+        try {
+            String namedMetric = Stream.of(tags)
+                    .filter((str -> StringUtils.isNotBlank(str) &&
+                            (str.contains("name") || str.contains("channel"))))
+                    .findAny()
+                    .orElse(null);
+            return namedMetric == null ?
```
lkemmerer: I think this might be more what you're looking for?
```
    String extractName(String[] tags) {
            return Stream.of(tags)
                    .filter(StringUtils::isNotBlank)
                    .filter(str -> str.contains("name") || str.contains("channel"))
                    .map(this::extractMetricName)
                    .findAny()
                    .orElse("");
    }

    private String extractMetricName(String fullMetricName) {
        try {
            return namedMetric.substring(namedMetric.indexOf(":") + 1);
        } catch (Exception ex) {
            return "";
        }
    }
```
That private method might need a better name, though.

src/main/java/com/flightstats/hub/metrics/StatsdReporter.java
```diff
@@ -83,9 +106,7 @@ public void time(String channel, String name, long start, long bytes, String...
     }
 
     public void mute() {
-        if (dataDogHandler != null) {
```
lkemmerer: It always scares me when random bits of code check for null values on injected members... 👍

src/test/java/com/flightstats/hub/metrics/StatsDFilterTest.java
```diff
@@ -151,4 +151,67 @@ public void testStatsdFilterIsSecondaryReporting_handleNullAndTrue() {
         StatsDFilter statsDFilter = new StatsDFilter(metricsConfig, channelConfigDao, webhookDao);
         assertTrue(statsDFilter.isSecondaryReporting("foo"));
     }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void testStatsdFilterExtractName_handleNullTags() {
```
lkemmerer: I don't know if it's normal, but I usually favor using checked exceptions for test methods rather than suppressing unchecked ones.

src/main/java/com/flightstats/hub/metrics/StatsDFilter.java
Paul-Hess: There is a specific design reason that I rewrote it this way:
  In the previous fashion, being that it is a black box whether the string we are comparing is a channel name or a webhook name or (very rarely now) neither both the the channel dynamo table and the webhook dynamo table could be queried before resolving the boolean. By doing it inline and putting the webhook evaluation second, if the channel matches, the extra call to the webhook dao (and the webhook cache and possibly the webhook dynamo table, which is set at half the throughput of the channel table) will not happen, saving a decent amount of overhead. Otherwise, I 100% agree that it was more readable and clean before.

All DAO classes implement the generic Dao interface :(

src/test/java/com/flightstats/hub/metrics/StatsDFilterTest.java
Paul-Hess: the cast of the LowerCasedCacheDaoDao<ChannelConfig> and CacheDao<webhook> (mocked) to their interface Dao implementation causes an unchecked warning, totally safe in this context, but I'm happy to change it for consistency's sake.

src/main/java/com/flightstats/hub/metrics/StatsDFilter.java
lkemmerer: The short-circuiting makes sense! One way to do that rather than with an or operator (which was the surprising/unintuitive part of that statement, I think) would be to drop the first `.orElse(false)` and do an `.orElse(splat-all-the-code-from-channel-checking-here.orElse(false))`. That way all the behavior is Java 8 Optional stuff, but it's nested and it's still kinda bleh.

That said, I think your idea of extracting the methods to get a balance of readability and performance is great. 👍

src/test/java/com/flightstats/hub/metrics/StatsDFilterTest.java
lkemmerer: Ooh, yeah. this is fine. I guess you could hide that ugliness in some methods (`getWebhookDao` and `getChannelConfigDao`) that suppress warnings and then the tests wouldn't have to worry about it.

Unrelatedly the imbalance of Channel*Config*Dao and WebhookDao (without the Config) really bugs me. :(

lkemmerer: I'm not sure my original suggestion actually made sense, tbh. It was a long day.

src/main/java/com/flightstats/hub/metrics/StatsDFilter.java
chriskessel: Maybe not relevant here, but there is Optional.orElseGet(Supplier) which can be nice if you need a short circuit. If you do something like Optional.of(thing).orElse(someMethod()), the someMethod() always gets called which is not necessarily intuitive. Well, it's obvious from a compiler viewpoint, but not from a human reading it viewpoint. I've gotten bitten by that a few times.

Sorry if that's old news or completely irrelevant.

lkemmerer: I forgot `.orElse` didn't short-circuit. Thanks!

src/test/java/com/flightstats/hub/metrics/StatsDFilterTest.java
Paul-Hess: > Unrelatedly the imbalance of ChannelConfigDao and WebhookDao (without the Config) really bugs me. :(

So much!

lkemmerer:
Might I suggest a squash + merge for this PR? 120 commits is a lot. :)

PR #1132, opened 2019-02-06
Update webhook LAST_COMPLETED ZK node with the minute before the first failed s3 verifier update

lkemmerer:
I resisted the urge to `stream`ify this logic, so (yay) I introduced more mutable state!

This will assume we update to the end range of the verifier run, but if it encounters an error, it'll set last completed to the minute before the failure. If that means that it got past the first minute, we'll update the ZK node. 

~Question: should we update it if the failure happens in the first minute after the start path (i.e we start the range at 12:00 and have a failure somewhere in the items at 12:01, should we update last completed to 12:00?)~ (LAST_COMPLETED minute is exclusive for spoke, so no need to do this)

### DIFF
--- a/src/main/java/com/flightstats/hub/dao/aws/S3Verifier.java
+++ b/src/main/java/com/flightstats/hub/dao/aws/S3Verifier.java
@@ -152,6 +152,7 @@ public class S3Verifier {
         String channelName = range.getChannelConfig().getDisplayName();
         SortedSet<ContentKey> keysToAdd = getMissing(range.getStartPath(), range.getEndPath(), channelName, s3SingleContentDao, new TreeSet<>());
         logger.debug("verifyChannel.starting {}", range);
+        MinutePath lastCompleted = range.getEndPath();
         for (ContentKey key : keysToAdd) {
             logger.trace("found missing {} {}", channelName, key);
             metricsService.increment(MISSING_ITEM_METRIC_NAME);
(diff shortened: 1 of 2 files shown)


src/main/java/com/flightstats/hub/dao/aws/S3Verifier.java
```diff
@@ -152,18 +152,25 @@ protected void verifyChannel(VerifierRange range) {
...
         for (ContentKey key : keysToAdd) {
             logger.trace("found missing {} {}", channelName, key);
             metricsService.increment(MISSING_ITEM_METRIC_NAME);
             boolean success = s3WriteQueue.add(new ChannelContentKey(channelName, key));
             if (!success) {
                 logger.error("unable to queue missing item {} {}", channelName, key);
                 metricsService.increment(VERIFIER_FAILED_METRIC_NAME);
-                return;
+                lastCompleted = new MinutePath(key.getTime().minusMinutes(1));
+                break;
```
Paul-Hess: I don't recall if lastCompleted as startPath becomes inclusive or exclusive in the range? If it is inclusive couldn't we set the last completed to key.getTime without the minus 1? 
Either way I think this will be helpful

lkemmerer: I don't recall either, but assumed it would return stuff starting with the beginning of the startPath minute. I'll check it out.

Paul-Hess: It is probably safer the way you did it anyway to avoid minute boundary issues now that I think about it

crozierm:
Look legit to me.

src/main/java/com/flightstats/hub/dao/aws/S3Verifier.java
lkemmerer: Follow up after Paul and I looked at the code:
- Spoke write DAO startPath is exclusive (nothing from start minute will be returned, so it assumes LAST_COMPLETED was successful)
- S3SingleContentDao startPath is inclusive (will return stuff from start minute.

So they aren't consistent. This is fine for now, I guess, since we're doing SPOKE KEYS - S3 KEYS, so if we get extra stuff from s3, we don't care. Means we're having to do more s3 queries, which is not great, but...whatever.

lkemmerer: So with that said, subtracting the minute from the failed item seems correct-ish.

PR #1133, opened 2019-02-11
S3 Verifier metrics cleanup and addition

lkemmerer:
Figured adding a couple more metrics wouldn't hurt, since grepping logs is a pain. Also a little bit of cleanup.

### DIFF
--- a/src/main/java/com/flightstats/hub/dao/aws/S3Verifier.java
+++ b/src/main/java/com/flightstats/hub/dao/aws/S3Verifier.java
@@ -146,7 +146,7 @@ public class S3Verifier {
     }
 
     @VisibleForTesting
-    protected void verifyChannel(VerifierRange range) {
+    void verifyChannel(VerifierRange range) {
         String channelName = range.getChannelConfig().getDisplayName();
         SortedSet<ContentKey> keysToAdd = getMissing(range.getStartPath(), range.getEndPath(), channelName, s3SingleContentDao, new TreeSet<>());
         logger.debug("verifyChannel.starting {}", range);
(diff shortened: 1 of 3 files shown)


src/main/java/com/flightstats/hub/metrics/PeriodicMetricEmitter.java
```diff
@@ -26,7 +25,6 @@ protected void runOneIteration() {
             metricsService.gauge("s3.writeQueue.threads", HubProperties.getS3WriteQueueThreads());
             metricsService.gauge("spoke.write.ttl", HubProperties.getSpokeTtlMinutes(SpokeStore.WRITE));
             metricsService.gauge("spoke.read.ttl", HubProperties.getSpokeTtlMinutes(SpokeStore.READ));
-            metricsService.count(S3Verifier.MISSING_ITEM_METRIC_NAME, 0);
```
lkemmerer: I was going to add the other metrics from the S3Verifier here, but Paul mentioned that statsd counts/increments reset on flush.

src/main/java/com/flightstats/hub/dao/aws/s3Verifier/VerifierMetrics.java
```diff
@@ -0,0 +1,17 @@
+package com.flightstats.hub.dao.aws.s3Verifier;
+
+public enum VerifierMetrics {
+    MISSING_ITEM("s3.verifier.missing"),
+    FAILED("s3.verifier.failed"),
+    TIMEOUT("s3.verifier.timeout");
+
```
manjula-chikkanagappa: This is looks so much better!!!

src/main/java/com/flightstats/hub/dao/aws/S3Verifier.java
```diff
@@ -146,7 +146,7 @@ VerifierRange getSingleVerifierRange(DateTime now, ChannelConfig channelConfig)
     }
 
     @VisibleForTesting
```
manjula-chikkanagappa: I still think this should be removed :-) 
As per this javadoc https://google.github.io/guava/releases/19.0/api/docs/com/google/common/annotations/VisibleForTesting.html
this annotation gives wide visibility to the member for testing purpose. Java access modifier is already taking care of the visibility. This is redundant.

lkemmerer: The annotation doesn't change accessibility of the method, though. It just notes that it is more widely accessible than it needs to be...and it's like that only for testing purposes. It also has a side effect of IntelliJ no longer complaining that you're using the wrong accessibility level.

PR #1134, opened 2019-02-14
Convert a bunch of guava Optionals to java's Optionals

lkemmerer:
Hub uses guava `Optional`s pretty heavily. I found this out when I tried to write `Optional.empty()` and got an error. IntelliJ agrees that java native's `Optional` is superior and it helped me do some refactoring. There were a couple cases more complicated than replacing `absent` with `empty` (the most complicated of them was replacing `asSet`), but generally this was pretty trivial. IntelliJ also decided to expand a bunch of `import *`s.

I didn't replace everything; just until the code started compiling again without mention of guava's `Optional` in any of the files I touched. It's a start, though.

If folks think this is a relatively safe change to consider merging, I'll deploy it to dev and run tests there beforehand. Local integration tests were green.


src/main/java/com/flightstats/hub/channel/ChannelContentResource.java
```diff
@@ -16,10 +16,18 @@
...
+import com.flightstats.hub.model.Content;
+import com.flightstats.hub.model.ContentKey;
+import com.flightstats.hub.model.ContentPath;
+import com.flightstats.hub.model.DirectionQuery;
+import com.flightstats.hub.model.Epoch;
+import com.flightstats.hub.model.InsertedContentKey;
+import com.flightstats.hub.model.Location;
+import com.flightstats.hub.model.Order;
+import com.flightstats.hub.model.SecondPath;
+import com.flightstats.hub.model.TimeQuery;
```
manjula-chikkanagappa: This looks nice 👍

src/main/java/com/flightstats/hub/channel/ChannelLatestResource.java
```diff
@@ -5,14 +5,14 @@
...
 import com.flightstats.hub.util.TimeUtil;
-import com.google.common.base.Optional;
 
 import javax.ws.rs.*;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 import java.net.URI;
+import java.util.Optional;
```
manjula-chikkanagappa: Yes. If functionality is available in java lib, we should definitely use that! 💯

PR #1135, opened 2019-02-14
Metrics filter

Paul-Hess:
Yeah it's big, but more than half is changing the name of metricsService to statsdReporter and deleting unused class cruft.
This contains commits from #1127 so is opened against that until it goes to master

The real guts of it:
Refactors the "metrics service", separating some of it's concerns and updating how it is instantiated to how we like to do things, names it statsdReporter for our current naming conventions.

Adds the ability to filter clients in the statsdReporter and report to one or both of them.

Adds tests around all the new things.

breaks out the lifecycle logic, calls the lifecycles lifecycle, registers the lifecycles outside of their constructors


src/main/java/com/flightstats/hub/filter/MetricsRequestFilter.java
```diff
@@ -71,21 +71,19 @@ public static void finalStats() {
...
             } else if (tags.get("call").endsWith("/shutdown")) {
-                logger.info("call to shutdown, ignoring datadog time {}", time);
+                logger.info("call to shutdown, ignoring statsd time {}", time);
             } else {
                 String[] tagArray = getTagArray(tags);
-                logger.trace("DataDog data sent: {}", Arrays.toString(tagArray));
-                if (metricsService.shouldLog(channel)) {
-                    metricsService.time("request", requestState.getStart(), tagArray);
-                }
+                logger.trace("statsdReporter data sent: {}", Arrays.toString(tagArray));
```
Paul-Hess: note: the shouldLog logic is now handled internally by the reporter

src/main/java/com/flightstats/hub/metrics/DataDogHandler.java
```diff
@@ -0,0 +1,65 @@
+package com.flightstats.hub.metrics;
+
```
Paul-Hess: This was formerly part of metricsService (now ~ statsdReporter) I broke it out and refactored very little, it is not 1:1 copy paste but further refactoring, I felt was too much on this already large PR

src/main/java/com/flightstats/hub/metrics/DataDogWhitelist.java
```diff
@@ -0,0 +1,12 @@
+package com.flightstats.hub.metrics;
```
Paul-Hess: This is a placeholder essentially, the next PR will handle how we will really filter the clients and this will almost definitely disappear

src/main/java/com/flightstats/hub/metrics/DataDogWhitelistProvider.java
```diff
@@ -0,0 +1,21 @@
+package com.flightstats.hub.metrics;
+
```
Paul-Hess: Same as comment above

src/main/java/com/flightstats/hub/metrics/MetricsResource.java
```diff
@@ -11,26 +12,45 @@
...
     }
 
     @GET
     @Path("trace")
     @Produces(MediaType.TEXT_PLAIN)
     public Response trigger() {
-        return Response.ok(MetricsRunner.logFilesInfo()).build();
+        return Response.ok(logFilesInfo()).build();
+    }
+
```
Paul-Hess: This logic in logFilesInfo was dumped into MetricsRunner (which mostly became CustomMetricsReporter) but had nothing really to do with it and is only called by this class, I started to refactor it and thought better of it at this time

src/main/java/com/flightstats/hub/metrics/StatsDFilter.java
```diff
@@ -0,0 +1,55 @@
+package com.flightstats.hub.metrics;
+
```
Paul-Hess: Another note that the actual filtering logic will change, but is in the scope of another PR.

src/test/java/com/flightstats/hub/webhook/WebhookErrorServiceUnitTest.java
```diff
@@ -1,9 +1,11 @@
 package com.flightstats.hub.webhook;
 
 import com.flightstats.hub.dao.ChannelService;
+import com.flightstats.hub.util.TimeUtil;
 import com.flightstats.hub.webhook.error.WebhookError;
 import com.flightstats.hub.webhook.error.WebhookErrorPruner;
 import com.flightstats.hub.webhook.error.WebhookErrorRepository;
+import org.joda.time.DateTime;
```
manjula-chikkanagappa: me and Laurie were thinking about using as features available in java platform as much as possible. This is one of those. Joda time was very popular before Java 8 . Now with Java 8 having good support for date,  we should be able to leverage that.

src/main/java/com/flightstats/hub/webhook/WebhookErrorService.java
```diff
@@ -61,7 +62,9 @@ public void delete(String webhook) {
     }
 
     private List<WebhookError> trimAndLookup(String webhook) {
-        List<WebhookError> errors = webhookErrorRepository.getErrors(webhook);
+        List<WebhookError> errors = webhookErrorRepository.getErrors(webhook).stream()
+                .sorted(Comparator.comparing(WebhookError::getCreationTime))
+                .collect(toList());
```
manjula-chikkanagappa: 👍

src/main/java/com/flightstats/hub/metrics/DataDogClient.java
```diff
@@ -0,0 +1,67 @@
...
+
+        if( "".equals(api_key) || "".equals(app_key)) {
+            logger.warn("datadog api_key or app_key not defined");
+            return;
+        }
+        String template = "{ \"message\": \"restarting\", \"scope\": \"name:%s\", \"end\": %d }";
+        try {
+
+            String data = String.format(template, name, fourMinutesFutureInSeconds);
+            String url = "https://app.datadoghq.com/api/v1/downtime?api_key=" +
```
manjula-chikkanagappa: Would it possible to externalize this property? It would be easier to change if this url ever changes.

src/test/javascript/integration/webhook_lifecycle_previous_spec.js
```diff
@@ -39,7 +39,7 @@ const callbackItems = [];
  * 2 - create a webhook on that channel with startItem=previous
  * 3 - start a server at the endpoint
  * 4 - post item into the channel
- * 5 - verify that the item are returned within delta time, incuding the second item posted in 2.
+ * 5 - verify that the item are returned within delta time, including the second item posted in 2.
```
manjula-chikkanagappa: I remember you mentioning that it would be easier to have all our tests in one language. Would this be part of refactor we do or do we create separate Jira ticket for this?

Paul-Hess: Yes, generally I think when refactoring a test it should be refactored to java, these tests however have nothing to do with this PR and refactoring them is well out of it's scope. I accidentally included these typo fixes in a commit and will remove it for now

src/test/java/com/flightstats/hub/webhook/WebhookErrorServiceUnitTest.java
Paul-Hess: You may have come across this while looking at the commit by commit view. This change was part of a previous PR that is already in master at this point. It showed up in a commit view possibly due to a pull from master, not sure, but it is not part of this PR.

src/main/java/com/flightstats/hub/metrics/DataDogClient.java
Paul-Hess: no problem, done

lkemmerer:
Is it best to look at the whole diff at once, or to go commit-by-commit?

Paul-Hess:
I think you're better off doing the whole diff it once

src/main/java/com/flightstats/hub/app/HubBindings.java
```diff
@@ -186,12 +187,22 @@ protected void configure() {
...
         bind(InfluxdbReporterLifecycle.class).asEagerSingleton();
 
+        bind(DataDogWhitelist.class).toProvider(DataDogWhitelistProvider.class).asEagerSingleton();
+        bind(StatsDFilter.class).asEagerSingleton();
+        bind(StatsdReporter.class).toProvider(StatsDReporterProvider.class).asEagerSingleton();
+        bind(StatsDReporterLifecycle.class).asEagerSingleton();
+        bind(CustomMetricsLifecycle.class).asEagerSingleton();
+        bind(PeriodicMetricEmitter.class).asEagerSingleton();
+        bind(PeriodicMetricEmitterLifecycle.class).asEagerSingleton();
+        ////
```
lkemmerer: super nit-picky, but you can get rid of this line.

src/main/java/com/flightstats/hub/app/HubMain.java
```diff
@@ -144,14 +149,23 @@ public Server startServer() throws Exception {
         return server;
     }
 
-    List<AbstractIdleService> getBeforeHealthCheckServices(Injector injector) {
-        List<AbstractIdleService> beforeHealthCheckServices = new ArrayList<>();
-        InfluxdbReporterLifecycle influxdbReporterLifecycle = injector.getInstance(InfluxdbReporterLifecycle.class);
-        beforeHealthCheckServices.add(influxdbReporterLifecycle);
-        return beforeHealthCheckServices;
+    private List<Service> getBeforeHealthCheckServices(Injector injector) {
+        return Stream.of(
```
lkemmerer: yay, nice.

src/main/java/com/flightstats/hub/app/ShutdownManager.java
```diff
@@ -51,7 +56,9 @@ protected void shutDown() throws Exception {
 
     public boolean shutdown(boolean useLock) throws Exception {
         logger.warn("shutting down!");
-        getMetricsService().mute();
+        String[] tags = { "restart", "shutdown" };
+        statsdReporter.event("Hub Restart Shutdown", "shutting down", tags);
```
lkemmerer: Wait, it used to mute metrics and then try to report that it was shutting down on that service?

src/main/java/com/flightstats/hub/metrics/ChannelType.java
```diff
@@ -0,0 +1,7 @@
+package com.flightstats.hub.metrics;
+
+public enum ChannelType {
```
lkemmerer: not 100% sure, but I think we generally use SCREAMING_SNAKE_CASE for enum values, since they're kinda like constants.

src/main/java/com/flightstats/hub/metrics/CustomMetricsLifecycle.java
```diff
@@ -0,0 +1,29 @@
+package com.flightstats.hub.metrics;
+
+import com.google.common.util.concurrent.AbstractScheduledService;
+import com.google.inject.Inject;
+
+import java.util.concurrent.TimeUnit;
+
+public class CustomMetricsLifecycle extends AbstractScheduledService{
+    private CustomMetricsReporter collector;
```
lkemmerer: I like this much better than the nested Service classes we have everywhere else. 👍

src/main/java/com/flightstats/hub/dao/aws/S3LargeContentDao.java
```diff
@@ -270,15 +270,15 @@ public void delete(String channel) {
     }
 
     public static class S3LargeContentDaoBuilder {
```
lkemmerer: Aw, this kind of makes me sad. We usually reserve builders for plain old domain objects, not stuff with logic in it. On top of that, I think this could be lombok-ified, but that would just make the outer class even weirder.

src/main/java/com/flightstats/hub/metrics/StatsdReporter.java
```diff
@@ -0,0 +1,91 @@
+package com.flightstats.hub.metrics;
+
+import com.google.inject.Inject;
+import com.timgroup.statsd.Event;
+import com.timgroup.statsd.StatsDClient;
+
+import java.util.List;
+import java.util.function.Consumer;
+
+public class StatsdReporter {
```
lkemmerer: Waaay earlier, I almost asked why you didn't just call it a plain old `MetricsReporter`. While that would have been an ok thing to do, I think you were right in leaving it more concrete for now.

My idle dream of having more generic, pluggable metrics reporting is pretty far down my list of nice-to-haves and it would introduce more complexity than we need. The fact that I think it'd be easy to convert over if ever we have time probably indicates that you did the right sort and amount of naming and refactoring.

src/test/java/com/flightstats/hub/metrics/DataDogHandlerIntegrationTest.java
```diff
@@ -0,0 +1,109 @@
...
+                .bindPort(8888)
+                .path("/api")
+                .build()
+                .httpServer();
+        httpServer.start();
+    }
+
+    @Test
+    public void testDatadogMute_mute() throws IOException {
+        MetricsConfig metricsConfig = MetricsConfig.builder()
```
lkemmerer: This is a heckuva test to figure out how to write. Nice job.

src/test/java/com/flightstats/hub/metrics/StatsDFormatterTest.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Test
+    public void testBuildCustomEvent_Error() {
+        MetricsConfig metricsConfig = MetricsConfig
+                .builder()
+                .build();
+        StatsDFormatter statsDFormatter = new StatsDFormatter(metricsConfig);
+        try {
+            statsDFormatter.buildCustomEvent("", "");
```
lkemmerer: I think there might be an `assertThrows` or some other assertion that'll handle testing that errors get thrown. This will silently pass if no exception is thrown.

src/main/java/com/flightstats/hub/app/ShutdownManager.java
Paul-Hess: Yep, that's exactly what it seems was happening.
The mute method hits the api endpoint at datadog to sleep monitoring on the host for a few minutes during restarts
We don't specify a start time in out api request so it would do the default and immediately mute, making the later call to event useless :)

https://docs.datadoghq.com/api/?lang=python#schedule-monitor-downtime

src/main/java/com/flightstats/hub/dao/aws/S3LargeContentDao.java
Paul-Hess: wait til you see ChannelConfig :(:(:(

Paul-Hess: agreed :(

src/test/java/com/flightstats/hub/metrics/StatsDFormatterTest.java
Paul-Hess: good catch, the second of these was silently failing... derp

src/main/java/com/flightstats/hub/metrics/StatsdReporter.java
Paul-Hess: 👍

src/main/java/com/flightstats/hub/metrics/ChannelType.java
Paul-Hess: good point, changed this.

src/main/java/com/flightstats/hub/app/ShutdownManager.java
lkemmerer: Aha. Thanks for fixing it. :)

PR #1138, opened 2019-02-19
PDXDDT-119: Add TTL Config for non current version objects in s3 bucket

manjula-chikkanagappa:


lkemmerer:
I am not very familiar with this part of the code, but from the poking around that I did, this makes sense and looks good. Thanks for testing it out!

PR #1139, opened 2019-02-20
use optionals return values for the channel config getters

Paul-Hess:


src/main/java/com/flightstats/hub/channel/ChannelResource.java
```diff
@@ -73,13 +73,14 @@
     public Response getChannelMetadata(@PathParam("channel") String channelName,
                                        @QueryParam("cached") @DefaultValue("true") boolean cached) {
         logger.debug("get channel {}", channelName);
-        ChannelConfig config = channelService.getChannelConfig(channelName, cached);
-        if (config == null) {
+
+        Optional<ChannelConfig> optionalChannelConfig = channelService.getChannelConfig(channelName, cached);
+        if (!optionalChannelConfig.isPresent()) {
```
lkemmerer: If you wanted to get all fancy and maybe less readable, you could do something like
```
if (!optionalChannelConfig.isPresent()) {
    logger.info("unable to get channel " + channelName);
}

return optionalChannelConfig
    .map(config -> buildChannelConfigResponse(config, uriInfo, channelName))
    .map(Response::ok)
    .map(ResponseBuilder(orwhateverthisis?)::build)
    .orElseThrow(...);
```
The logging kinda messes up the flow a little bit, but also that's a lot of maps. I honestly hate calling `Optional.get` though.

lkemmerer: Oh wait, you might be able to do `orElseThrow(() -> { logger.info(...); return new WebApplicationException(...);)` Up to you. That's also kinda weird.

src/main/java/com/flightstats/hub/channel/ChannelResource.java
```diff
@@ -253,11 +258,11 @@ public EventOutput getEvents(@PathParam("channel") String channel, @HeaderParam(
...
     @DELETE
     public Response delete(@PathParam("channel") final String channelName) throws Exception {
-        ChannelConfig channelConfig = channelService.getChannelConfig(channelName, false);
-        if (channelConfig == null) {
+        Optional<ChannelConfig> optionalChannelConfig = channelService.getChannelConfig(channelName, false);
+        if (!optionalChannelConfig.isPresent()) {
             return notFound(channelName);
         }
-        if (HubProperties.isProtected() || channelConfig.isProtect()) {
+        if (HubProperties.isProtected() || optionalChannelConfig.get().isProtect()) {
```
lkemmerer: I was going to suggest chaining this stuff up with some `map`s and `filter`s, but it's probably a little gross.

src/main/java/com/flightstats/hub/channel/ChannelStatusResource.java
```diff
@@ -31,10 +31,11 @@
     public Response getLatest(@PathParam("channel") String channel,
                               @QueryParam("stable") @DefaultValue("true") boolean stable,
                               @QueryParam("trace") @DefaultValue("false") boolean trace) {
-        ChannelConfig channelConfig = channelService.getCachedChannelConfig(channel);
-        if (null == channelConfig) {
+        Optional<ChannelConfig> optionalChannelConfig = channelService.getCachedChannelConfig(channel);
```
lkemmerer: I would probably extract the stuff from line 38 onwards into another method that takes a `ChannelConfig` and whatever else it needs, which'd make this a 3 or 4 line method, but... the rest of this method is not straightforward and probably needs more love than just that. So forget I said anything.

Paul-Hess:
I had already started some refactoring based on your comments yesterday. I can revert them if you think too much. I left some the way they were as refactoring them didn't add anything to all of them.

src/main/java/com/flightstats/hub/channel/ChannelResource.java
```diff
@@ -71,53 +70,53 @@
...
-                optionalChannelConfig.orElse(null),
+        ChannelConfig newChannelConfig = ChannelConfig.createFromJsonWithName(json, channelName);
+        
+        ChannelConfig oldChannelConfig = channelService
+                .getChannelConfig(channelName, false)
+                .orElse(newChannelConfig);
+
+
+        logger.info("creating channel {} {}", newChannelConfig, newChannelConfig.getCreationDate().getTime());
+        newChannelConfig = channelService.updateChannel(
```
lkemmerer: I'm not a fan of using `newChannelConfig` here because of mutability (even though it's scoped to the method). I'd create a new `channelConfig` or `updatedChannelConfig` or something instead.

src/main/java/com/flightstats/hub/webhook/TimedWebhookStrategy.java
```diff
@@ -187,8 +187,10 @@ private void doWork() throws InterruptedException {
                     nextTime = lastAdded.getTime();
                 }
                 DateTime stable = TimeUtil.stable().minus(duration);
-                Optional<ChannelConfig> optionalChannelConfig = channelService.getChannelConfig(channel, true);
-                if (optionalChannelConfig.isPresent() && !optionalChannelConfig.get().isLive()) {
+                boolean notLiveChannel = channelService.getChannelConfig(channel, true)
```
lkemmerer: I dunno if this'd be a worthwhile method to extract over to the channelService here, but we use it in a couple places, so it might be a nice bit of tidying up.

lkemmerer:
I like the extra refactoring and don't think you should revert it. :)

src/main/java/com/flightstats/hub/webhook/TimedWebhookStrategy.java
Paul-Hess: hmm, this is also backwards, it should filter OUT the "live" channel 👎 glad that you mentioned something

src/main/java/com/flightstats/hub/dao/ChannelService.java
```diff
@@ -289,6 +286,12 @@ public void deleteBefore(String channel, ContentKey limitKey) {
         return Optional.ofNullable(channelConfig);
     }
 
+    public boolean isNotLiveChannel(String channelName, boolean checkCache) {
```
lkemmerer: I'd probably have left this as `isLiveChannel` for readability but I bet IntelliJ thought you should invert it because it's only used in that state. 6 of one, half dozen of the other. 🤷‍♂️

src/main/java/com/flightstats/hub/channel/InternalChannelResource.java
```diff
@@ -16,10 +16,19 @@
...
+import javax.ws.rs.DELETE;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.DefaultValue;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.UriInfo;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriBuilder;
```
manjula-chikkanagappa: 👍

src/main/java/com/flightstats/hub/channel/ChannelResource.java
```diff
@@ -71,53 +70,53 @@
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     public Response getChannelMetadata(@PathParam("channel") String channelName,
-                                       @QueryParam("cached") @DefaultValue("true") boolean cached) {
```
manjula-chikkanagappa: I thought we are using @Sl4fj  as we refactor code.

Paul-Hess:
closing this as it is merged into #1127

PR #1140, opened 2019-02-21
Datadog filter channel

Paul-Hess:


Paul-Hess:
This will probably be the easiest to review, but it contains commits from 3 other PRs, may want to hold off from going through it until those are approved and not subject to change

src/main/java/com/flightstats/hub/metrics/StatsDFilter.java
```diff
@@ -1,18 +1,19 @@
...
 import com.timgroup.statsd.NoOpStatsDClient;
 import com.timgroup.statsd.NonBlockingStatsDClient;
 import com.timgroup.statsd.StatsDClient;
 
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
+import java.util.Optional;
```
manjula-chikkanagappa: 👍

src/test/java/com/flightstats/hub/webhook/WebhookTest.java
```diff
@@ -141,4 +147,12 @@ public void testIsTagPrototype() throws Exception {
         assertTrue(twh.isTagPrototype());
     }
 
+    @Test
+    public void testSecondaryMetricsReporting() {
+        Webhook withDefaults = this.webhook.withDefaults();
+        assertFalse(withDefaults.isSecondaryMetricsReporting());
+        String json = "{ \"secondaryMetricsReporting\": \"true\" }";
```
manjula-chikkanagappa: is true a string or boolean? if boolean, quotes not required.

src/main/java/com/flightstats/hub/model/ChannelConfig.java
```diff
@@ -46,10 +50,23 @@
...
+                          boolean keepForever,
+                          String description,
+                          Set<String> tags,
+                          String replicationSource,
+                          String storage,
+                          boolean protect,
+                          DateTime mutableTime,
+                          boolean allowZeroBytes,
+                          String displayName,
+                          String secondaryMetricsReporting) {
```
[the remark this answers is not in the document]
lkemmerer: I agree. This looks like one of the classes that got de-lombokified a while back. Assuming it doesn't have too much logic in it, I'd definitely convert it back.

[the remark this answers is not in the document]
lkemmerer: ...and if it does have too much logic in it, maybe that could be moved over to a different class if it's easy to extricate.

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
```diff
@@ -4,29 +4,25 @@
...
-import org.slf4j.LoggerFactory;
 
-import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 
 import static org.junit.Assert.*;
 
 public class ChannelConfigTest {
-    private static final Logger logger = LoggerFactory.getLogger(ChannelConfigTest.class);
```
lkemmerer: On a tangent, I am not 100% sure why we use a logger factory all over the place. Seems like there'd be some sort of annotation we could use.

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
```diff
@@ -137,13 +133,13 @@ public void testCopy() throws IOException {
                 .mutableTime(TimeUtil.now())
                 .allowZeroBytes(false)
                 .build();
-        assertTrue(config.equals(config.toBuilder().build()));
 
-        assertTrue(config.equals(ChannelConfig.createFromJson(config.toJson())));
+        assertEquals(config, config.toBuilder().build());
```
lkemmerer: ...I don't know what `.toBuilder()` does and I'm a little bit afraid to look. I know you didn't introduce it, so don't feel obliged to answer if you don't know.

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
```diff
@@ -189,4 +185,34 @@ public void testZeroBytes() {
         assertFalse(testZeroBytes.isAllowZeroBytes());
     }
 
+    @Test
+    public void testSecondaryMetricsReporting_defaultEmpty() {
+        ChannelConfig testSecondaryMetricsNull = ChannelConfig
+                .builder()
+                .name("testSecondaryMetrics0")
+                .build();
+        assertEquals("", testSecondaryMetricsNull.getSecondaryMetricsReporting());
```
lkemmerer: If you change this to a lombok object, you'll either need to add a default value on this or write your own getter. I guess either one isn't too bad...

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
chriskessel: toBuilder() is full of awesomeness. It creates an exact close of the original object, but returns a builder. Great for  "I need a clone, but with this one field changed" situations.

That said...no idea why you'd to a toBuilder() and then immediately .build().

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
lkemmerer: On the bright side, if you just do a default value, you can probably get rid of the next test down.

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
chriskessel: ```@Slf4j```

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
lkemmerer: Ah, neat! I usually use the lombok annotation that lets you do `withAttribute()` chaining to accomplish something similar. ...As you pointed out, I agree that the main concern was the immediate `build()` that happens afterwards.

lkemmerer:
Thanks for cleaning up those test assertions. :)

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
lkemmerer: Ooh. I was afraid the reason we were doing this was that we weren't using that library. I'm absolutely going to start converting these. Thanks!

chriskessel: I think it should work as there's some slf4j bridge to most loggers that seems to magically happen (presumably somehow via lombok or it's dependencies). You'd have to verify it though.

lkemmerer: There are a couple annotations in the codebase already (I think Zander started adding them before he left). I'll double check that they're working.

lkemmerer: Yep, they're working just fine.
As a side note, I'm going to start a confluence doc with common cleanup things we should try to do in the codebase as we touch stuff.

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
Paul-Hess: I am guessing the purpose of this was to test that the (all hand-rolled) hash and equals and builder and toBuilder worked in unison to create an exact copy. Which IIRC is implicitly tested on the next line as createFromJson uses that method under the hood.

src/test/java/com/flightstats/hub/webhook/WebhookTest.java
Paul-Hess: true :) pun intended the quotes are not necessary

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
Paul-Hess: 👍

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
lkemmerer: Ah, I guess it'd make sense to leave it there at least temporarily to make sure re-lombokifying things works out if you decide to make that change.

src/main/java/com/flightstats/hub/model/ChannelConfig.java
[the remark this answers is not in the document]
Paul-Hess: carving out this as a new PR against this one here https://github.com/flightstats/hub/tree/refactor-channel-config

PR #1141, opened 2019-02-22
Removed unused imports, format and rearrange code using IntelliJ form…

manjula-chikkanagappa:
…atter and update API doc


docs/pages/mydoc/hub_channels_insert.md
```diff
@@ -94,19 +94,19 @@ On success: `HTTP/1.1 201 Created`
...
-      "http://hub/channel/stumptown/2013/04/23/20/42/31/749/{hash}001"
-    ]
-  }
-}
+   "_links": {
+     "self": {
+       "href": "http://hub/channel/stumptown/2019/02/22/17/36/52/977/GQa3SX/next/2?stable=false"
+     },
+     "uris": [
+       "http://hub.dls.dev.flightstats.io/channel/stumptown/2019/02/22/17/36/52/977/GQa3SX000000",
```
lkemmerer: There are a couple missed domain -> `hub` replacements here.

src/main/java/com/flightstats/hub/app/ClusterHubBindings.java
```diff
@@ -28,6 +25,22 @@
 class ClusterHubBindings extends AbstractModule {
     private final static Logger logger = LoggerFactory.getLogger(ClusterHubBindings.class);
 
+    @Inject
```
lkemmerer: Were these automatically moved? I like having the `@Override`/`configure()` method first since I think it's the one special sort of method in this file and the others are individual dependency specifications.

src/main/java/com/flightstats/hub/app/HubBindings.java
```diff
@@ -15,13 +15,7 @@
...
 import com.flightstats.hub.replication.ReplicationManager;
 import com.flightstats.hub.rest.*;
-import com.flightstats.hub.spoke.FileSpokeStore;
-import com.flightstats.hub.spoke.GCRunner;
-import com.flightstats.hub.spoke.SpokeClusterRegister;
-import com.flightstats.hub.spoke.SpokeFinalCheck;
-import com.flightstats.hub.spoke.SpokeReadContentDao;
-import com.flightstats.hub.spoke.SpokeStore;
-import com.flightstats.hub.spoke.SpokeWriteContentDao;
+import com.flightstats.hub.spoke.*;
```
lkemmerer: The one non-default IntelliJ setting we have (in the hub code, anyway) is that we don't splat our imports (I have my limit set to 999 or something silly). I think I may have neglected to specify that the last time you asked about how we format our code. Sorry!

src/main/java/com/flightstats/hub/app/SingleHubBindings.java
```diff
@@ -44,4 +32,11 @@ protected void configure() {
     public static Dao<Webhook> buildWebhookDao(WatchManager watchManager, FileWebhookDao dao) {
         return new CachedDao<>(dao, watchManager, "/webhooks/cache");
     }
+
+    @Override
+    protected void configure() {
```
lkemmerer: Same comment here about the configure. I just realized it's probably because the others are `static`, but I guess I'm not a stickler for ordering by that so much as ordering by what I see as importance/ergonomics? Dunno. I guess it doesn't matter that much.

lkemmerer:
Other than the documentation change (where there's a flightstats url, and removing the splatted imports, this looks good. I think the overridden methods vs the statics ordering thing rubbed me the wrong way because I almost always try to avoid having classes that have both types of methods and it was just weird to have to think about.

src/main/java/com/flightstats/hub/app/HubBindings.java
manjula-chikkanagappa: OOPs! I do not like these imports at all. i will replace them .

crozierm:
Yes!

One thing... is this going to create merge conflicts with anything else that is in-flight? We definitely want to do this, but maybe we should get some other PRs through

manjula-chikkanagappa:
> Yes!
> 
> One thing... is this going to create merge conflicts with anything else that is in-flight? We definitely want to do this, but maybe we should get some other PRs through

Ok. Thanks. I will coordinate with Laurie and Paul on this one.

PR #1142, opened 2019-02-26
Stop periodically looking for empty webhook leader zk nodes, just delete them when we delete webhooks

lkemmerer:
As part of the zookeeper cleanup, I created a job that would go through and find webhook leader nodes without locks and leases. This executed the same code that had originally been running whenever a node started up.

It turned out the code doesn't do what I thought it did. After reading the logs, it looks like due to the lifecycle of ZK webhook leader locks, sometimes an active webhook meets the requirements in the existing code and the node was getting deleted (without releasing leadership) and then recreated shortly afterwards, occasionally causing two nodes to think they were leaders.

So I reverted the periodic cleanup and the strange code will go back to only executing on startup (Commit 1). In commit 2, I solve the same problem in a (hopefully) safer way, by adding webhook leader ZK node deletion to part of the process of deleting webhook state when we delete webhooks...

I think this fixes the main issue that was causing the frequent duplicate webhook item issue that was introduced earlier. The work I have in another branch around curator locks and putting a lock around the WebhookLeader node will (probably) sew up the other duplicate issues that had been seen previously.

### DIFF
--- a/src/main/java/com/flightstats/hub/webhook/WebhookErrorService.java
+++ b/src/main/java/com/flightstats/hub/webhook/WebhookErrorService.java
@@ -10,8 +10,7 @@ import com.flightstats.hub.webhook.error.WebhookErrorPruner;
 import com.flightstats.hub.webhook.error.WebhookErrorRepository;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import lombok.extern.slf4j.Slf4j;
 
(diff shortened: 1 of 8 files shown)


chriskessel:
:heart:

src/main/java/com/flightstats/hub/webhook/WebhookErrorService.java
```diff
@@ -31,9 +30,8 @@
  */
 
 @Singleton
+@Slf4j
```
Paul-Hess: 👍

PR #1143, opened 2019-02-27
close s3 object connection after read

manjula-chikkanagappa:


src/main/java/com/flightstats/hub/dao/aws/S3LargeContentDao.java
```diff
@@ -184,26 +182,26 @@ public Content get(final String channelName, final ContentKey key) {
...
     }
 
     private Content getS3Object(String channelName, ContentKey key) throws IOException {
+
         long start = System.currentTimeMillis();
-        try {
-            GetObjectRequest request = new GetObjectRequest(s3BucketName.getS3BucketName(), getS3ContentKey(channelName, key, false));
-            S3Object object = s3Client.getObject(request);
+        GetObjectRequest request = new GetObjectRequest(s3BucketName.getS3BucketName(), getS3ContentKey(channelName, key, false));
+        try (S3Object object = s3Client.getObject(request)) {
```
lkemmerer: I'm not entirely sure this is going to do what you expect it to. According to the docs (https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/S3Object.html#getObjectContent-- and https://github.com/aws/aws-sdk-java/blob/master/src/samples/AmazonS3/S3Sample.java), the connection will remain open until the stream (`getObjectContent`) is read to completion or explicitly closed. Since this method just copies the stream into a `Content` object, I think we'll need to make sure all the callers that read the stream have a finally block that either calls `Content#getObjectContent().close()` or `Content#close()`. If possible, I'd write some tests to confirm.

lkemmerer:
Thanks for doing all that cleanup!

src/main/java/com/flightstats/hub/dao/aws/S3LargeContentDao.java
manjula-chikkanagappa: Yes correct. The integration tests passed locally on single instance. These tests failing in cluster mode.

PR #1144, opened 2019-02-27
PDXDDT-677 Remove special S3 TTL functionality when maxRules=0

manjula-chikkanagappa:


src/main/java/com/flightstats/hub/dao/aws/S3Config.java
```diff
@@ -17,15 +17,21 @@
...
+import java.util.Random;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
 
+@Slf4j
 public class S3Config {
-    private final static Logger logger = LoggerFactory.getLogger(S3Config.class);
+
+    private static final Integer S3_LIFECYCLE_RULES_MAX = 1000;
```
lkemmerer: Might I suggest you set this to 990 or 999 so that by default it doesn't interfere with the terraform rule we'd like to have? I suppose if we use one of those, we'd either need to change the name or add an explanatory comment, though.

manjula-chikkanagappa: I thought about it. I was not sure number of rules we will adding  in terraform. Given the scarcity of rules I did not want to pick a number :)

src/main/java/com/flightstats/hub/dao/aws/S3Config.java
```diff
@@ -130,26 +137,23 @@ private void updateMaxItems(ChannelConfig config, ContentKey latest) {
...
+            int maxRules = HubProperties.getProperty("s3.maxRules", S3_LIFECYCLE_RULES_AVAILABLE);
             List<BucketLifecycleConfiguration.Rule> rules = new ArrayList<>();
-            if (maxRules == 0) {
-                rules.add(new BucketLifecycleConfiguration.Rule()
-                        .withId("OneDay")
-                        .withExpirationInDays(1)
-                        .withStatus(BucketLifecycleConfiguration.ENABLED));
-            } else {
+
+            if (maxRules > 0 && maxRules <= S3_LIFECYCLE_RULES_AVAILABLE) {
```
Paul-Hess: I don't have the best understanding of these rule configurations, so just asking:
If this statement is false, we do nothing, what happens then? We have no rules apportioned? Or is there an automatic default?

manjula-chikkanagappa: if s3.maxRules property is not available, default value is taken (990)

PR #1145, opened 2019-03-05
Curator lock refactoring

lkemmerer:
This fixes the InterProcessorSemaphoreMutex state issue we found. It is mostly renaming stuff, extracting things, and writing tests, though. I'm up for better name suggestions.  It also adds a synchronous/blocking version of the curator lock (now called Distributed(Asynchronous)LockRunner) because the use case the WebhookManager has doesn't require a long-running process and the extra complexity is silly to add for no reason.

Going commit-by-commit is probably more digestible because of renaming. This has been sitting unchanged in a branch for a while, so hopefully I'll have answers to questions.


src/main/java/com/flightstats/hub/cluster/DistributedAsynchronousLockRunner.java
```diff
@@ -1,98 +1,62 @@
...
  * DistributedAsynchronousLockRunner is intended for long or short running processes.
  */
+@Slf4j
 public class DistributedAsynchronousLockRunner {
-    private final static Logger logger = LoggerFactory.getLogger(DistributedAsynchronousLockRunner.class);
-
-    private final CuratorFramework curator;
-    private final ExecutorService singleThreadExecutor;
-    private final Leadership leadership;
+    private final DistributedLeadershipLockManager leadershipLockManager;
```
manjula-chikkanagappa: Its a long name. Not sure if we can creatively shorten this. But is should communicate its intent clearly which it is now :)

manjula-chikkanagappa: can we just say Leader instead of leadership !

lkemmerer: Yeah... I definitely didn't know what a `CuratorLock` was off-hand and it wasn't clear that it was non-blocking. I couldn't think of something better and I tend to err on the side of super long names in those cases.

src/main/java/com/flightstats/hub/cluster/DistributedLeadershipLockManager.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.recipes.locks.InterProcessSemaphoreMutex;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import java.util.concurrent.TimeUnit;
+
+public class DistributedLeadershipLockManager {
+    private final static Logger logger = LoggerFactory.getLogger(DistributedLeadershipLockManager.class);
```
manjula-chikkanagappa: @Slf4j? Since we started replacing log statements with lombok annotation as part of refactor, we shud continue doing it.

src/main/java/com/flightstats/hub/cluster/DistributedAsynchronousLockRunner.java
lkemmerer: Yeah, I can do that! And whoops, I thought the name you were referring to was the LockRunner.

src/main/java/com/flightstats/hub/cluster/DistributedLeadershipLockManager.java
lkemmerer: Whoops, I thought I went back and changed all those. 👍

src/main/java/com/flightstats/hub/dao/aws/S3Verifier.java
```diff
@@ -67,17 +65,15 @@ public S3Verifier(LastContentPath lastContentPath,
...
         this.channelService = channelService;
         this.spokeWriteContentDao = spokeWriteContentDao;
         this.s3SingleContentDao = s3SingleContentDao;
         this.s3WriteQueue = s3WriteQueue;
         this.httpClient = httpClient;
-        this.zooKeeperState = zooKeeperState;
-        this.curator = curator;
+        this.distributedLeadershipLockManager = distributedLeadershipLockManager;
         this.metricsService = metricsService;
         this.baseTimeoutMinutes = HubProperties.getProperty("s3Verifier.baseTimeoutMinutes", 2);
```
manjula-chikkanagappa: This class needs refactoring sometime ! Too many arguments :)

src/test/java/com/flightstats/hub/cluster/DistributedAsynchronousLockRunnerTest.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class DistributedAsynchronousLockRunnerTest {
```
manjula-chikkanagappa: Same here. can we shorten the corresponding class name to DistributedAsyncLockTest?

src/test/java/com/flightstats/hub/cluster/DistributedAsynchronousLockRunnerTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+        assertTrue(runWithLock.isPresent());
+
+        latch.await(1, TimeUnit.MINUTES);
+
+        assertEquals(newArrayList("lockable1"), lockList.get());
+    }
+
+    @Test
+    public void testPreventsTwoLockablesFromRunningAtTheSameTime() throws Exception {
```
manjula-chikkanagappa: can we break name with underscores for more readability? something like like this https://www.baeldung.com/integration-testing-a-rest-api. I know underscores in name is not a java standard. But !

src/test/java/com/flightstats/hub/dao/aws/S3VerifierUnitTest.java
```diff
@@ -1,7 +1,7 @@
 package com.flightstats.hub.dao.aws;
```
manjula-chikkanagappa: It would be nice to remove unit from class name. Eventually all integrationtests will be in integrationTest folder which i am currently working on.

src/main/java/com/flightstats/hub/cluster/Leadership.java
```diff
@@ -1,12 +1,11 @@
 package com.flightstats.hub.cluster;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import lombok.extern.slf4j.Slf4j;
 
 import java.util.concurrent.atomic.AtomicBoolean;
 
+@Slf4j
```
manjula-chikkanagappa: 👍

src/main/java/com/flightstats/hub/cluster/DistributedLockRunner.java
```diff
@@ -0,0 +1,39 @@
...
+
+import javax.inject.Inject;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+
+/**
+ * DistributedLockRunner is intended for short running processes, since it is blocking.
+ */
+@Slf4j
+public class DistributedLockRunner {
```
manjula-chikkanagappa: same here. Why is it LockRunner? Why not just say DistibutedLock?

src/main/java/com/flightstats/hub/dao/aws/S3Verifier.java
lkemmerer: I have a branch that does a little of that, though I think it extracts things and adds more dependencies.... (https://github.com/flightstats/hub/pull/1129/files) Maybe I'll catch up with stuff at some point and get that cleaned up a little more.

manjula-chikkanagappa: Awesome to see all the refactoring going in to this code. Cannot wait for the day when we are ready to integrate with Kafka :)

src/test/java/com/flightstats/hub/cluster/DistributedAsynchronousLockRunnerTest.java
lkemmerer: In other project I have the habit of doing `test_methodName_assertionHere_givenWhatever` type names... so something like 'test_absoluteValue_returnsAPositiveNumber_givenANegativeNumber` I think keeping the "when"/method name near the start of the test name is helpful for scanning files and grouping, but I'm cool with whatever order of given/then feels good.

src/test/java/com/flightstats/hub/dao/aws/S3VerifierUnitTest.java
lkemmerer: I think I'll tackle that once  we have that package/directory. I'd also like to keep integration tests (testing groups of classes) separate from functional/system tests, which I think of as being more black-box. That way we can draw better lines about what dependencies we have access to in tests. I think the system tests should only be able to access the hub client and its models but shouldn't access most of the hub code itself.

src/main/java/com/flightstats/hub/cluster/DistributedLockRunner.java
lkemmerer: I was thinking that it's the class that runs a thing in a lock, not the lock itself...

PR #1146, opened 2019-03-07
PDXDDT-679 Java system test for webhooks

manjula-chikkanagappa:
I have added a system test for testing webhook lifecycle. This test creates callback server programmatically and tears down once the tests are done. I had run into some synchronization issues during development. This test now is pretty stable.
I have updated README no how to tun system tests on local setup and Jenkins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,3 +34,24 @@ The hub is a long term data store.
 Run ./gradlew build
 
 The above command compiles the code, runs the unit tests and integration tests.~~~~
+
+
+### How to run System tests:
+
+From Local:
+
+1) Setup hub:
+    docker run -p 80:80 flightstats/hub:latest
+   
+2) set base.url property in intergation-hub.properties
+    
+3) Go to root of hub project source and run below command
+   ./gradlew systemTest -i    
+   
+From Jenkins:
+
+1) Goto https://ddt-jenkins.pdx.prod.flightstats.io/job/hub-system-test/ 
+2) Click on "Build With Parameters" 
+3) Add your git branch and start the build.
+
+  
--- a/README.md
+++ b/README.md
@@ -41,7 +41,8 @@ The above command compiles the code, runs the unit tests and integration tests.~
 From Local:
 
 1) Setup hub:
-    docker run -p 80:80 flightstats/hub:latest
+    a) ./localDocker.sh (from root directory)
+    b) docker run -p 80:80 hub:local
    
 2) set base.url property in intergation-hub.properties
     
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 apply plugin: 'java'
-apply plugin: 'idea'
+
 apply plugin: 'application'
 
 if (project.ext.properties.get('isDevBuild') == null) {
@@ -8,6 +8,22 @@ if (project.ext.properties.get('isDevBuild') == null) {
 
 repositories {
     mavenCentral()
+}
+
+sourceSets {
+    systemTest {
+        java {
+            compileClasspath += main.output + test.output
+            runtimeClasspath += main.output + test.output
+            srcDir file('src/systemTest/java')
+        }
+        resources.srcDir file('src/systemTest/resources')
+    }
+}
+
+configurations {
+    systemTestCompile.extendsFrom testCompile
+    systemTestRuntime.extendsFrom testRuntime
 }
 
 sourceCompatibility = JavaVersion.VERSION_1_8
@@ -31,6 +47,12 @@ mainClassName = "com.flightstats.hub.app.SingleHubMain"
 task nasTest(type: Test) {
     exclude 'com/flightstats/hub/dao/aws/**'
     exclude 'com/flightstats/hub/spoke/**'
+}
+
+task systemTest(type: Test, description: 'Runs the system tests.', group: 'Verification') {
+    testClassesDirs = sourceSets.systemTest.output.classesDirs
+    classpath = sourceSets.systemTest.runtimeClasspath
+    mustRunAfter tasks.test
 }
 
 wrapper {
@@ -108,7 +130,7 @@ dependencies {
 
     compile 'com.atlassian.commonmark:commonmark:0.10.0'
 
-    compileOnly 'org.projectlombok:lombok:1.18.6'
+    compile 'org.projectlombok:lombok:1.18.6'
     annotationProcessor 'org.projectlombok:lombok:1.18.6'
 
     testCompileOnly 'org.projectlombok:lombok:1.18.6'
@@ -116,6 +138,21 @@ dependencies {
     testCompile 'com.sparkjava:spark-core:2.7.1'
     testCompile 'junit:junit:4.+'
     testCompile "org.mockito:mockito-all:1.+"
+
+    systemTestAnnotationProcessor 'org.projectlombok:lombok:1.18.6'
+
+    systemTestCompile 'org.eclipse.jetty:jetty-server:9.4.15.v20190215'
+    systemTestCompile 'org.eclipse.jetty:jetty-servlet:9.4.15.v20190215'
+
+    systemTestCompile 'com.squareup.retrofit2:retrofit:2.4.0'
+    systemTestCompile 'com.squareup.retrofit2:converter-gson:2.5.0'
+    systemTestCompile 'com.squareup.retrofit2:converter-scalars:2.5.0'
+
+    systemTestCompile 'org.awaitility:awaitility:3.0.0'
+    systemTestCompile 'org.awaitility:awaitility-proxy:3.1.6'
+
+    systemTestCompile 'org.glassfish.jersey.containers:jersey-container-servlet-core:2.28'
+    systemTestCompile 'org.glassfish.jersey.media:jersey-media-json-jackson:2.28'
 
 }
 
--- a/src/systemTest/java/com/flightstats/hub/callback/CallbackCache.java
+++ b/src/systemTest/java/com/flightstats/hub/callback/CallbackCache.java
@@ -1,0 +1,37 @@
+package com.flightstats.hub.callback;
+
+import com.flightstats.hub.model.WebhookCallbackRequest;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+@Slf4j
+public class CallbackCache {
+
+    private final Map<String, List<String>> cacheRequestObject = new ConcurrentHashMap<>();
+
+    public void put(WebhookCallbackRequest webhookCallbackRequest) {
+        String webhookName = webhookCallbackRequest.getName();
+
+        synchronized (this) {
+            if (contains(webhookName)) {
+                get(webhookName).addAll(webhookCallbackRequest.getUris());
+            } else {
+                cacheRequestObject.put(
+                        webhookCallbackRequest.getName(),
+                        new ArrayList<>(webhookCallbackRequest.getUris()));
+            }
+        }
+    }
+
+    public List<String> get(String webhookName) {
+        return cacheRequestObject.get(webhookName);
+    }
+
+    public boolean contains(String webhookName) {
+        return cacheRequestObject.containsKey(webhookName);
+    }
+}
--- a/src/systemTest/java/com/flightstats/hub/callback/CallbackResource.java
+++ b/src/systemTest/java/com/flightstats/hub/callback/CallbackResource.java
@@ -1,0 +1,54 @@
+package com.flightstats.hub.callback;
+
+import com.flightstats.hub.model.WebhookCallbackRequest;
+import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
+
+import javax.inject.Inject;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+@Slf4j
+@Path("/callback")
+public class CallbackResource {
+
+    private CallbackCache callbackCache;
+
+    @Inject
+    public CallbackResource(CallbackCache callbackCache) {
+        this.callbackCache = callbackCache;
+    }
+
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @SneakyThrows
+    public Response create(WebhookCallbackRequest webhookCallbackRequest) {
+        log.info("Callback request received {} ", webhookCallbackRequest);
+        this.callbackCache.put(webhookCallbackRequest);
+        return Response.status(Response.Status.OK).build();
+    }
+
+    @GET
+    @Path("/{webhookName}")
+    @Produces(MediaType.TEXT_PLAIN)
+    @SneakyThrows
+    public Response get(@PathParam("webhookName") String webhookName) {
+
+        if (this.callbackCache.contains(webhookName)) {
+            return Response
+                    .ok(this.callbackCache.get(webhookName).toString())
+                    .build();
+        } else {
+            return Response
+                    .status(Response.Status.NOT_FOUND)
+                    .build();
+        }
+    }
+
+}
--- a/src/systemTest/java/com/flightstats/hub/callback/CallbackServer.java
+++ b/src/systemTest/java/com/flightstats/hub/callback/CallbackServer.java
@@ -1,0 +1,50 @@
+package com.flightstats.hub.callback;
+
+import lombok.SneakyThrows;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.servlet.ServletContextHandler;
+import org.eclipse.jetty.servlet.ServletHolder;
+import org.glassfish.jersey.servlet.ServletContainer;
+
+import java.net.InetAddress;
+
+public class CallbackServer {
+
+    private static final int JETTY_PORT = 8090;
+    private Server jettyServer;
+
+    @SneakyThrows
+    public void start() {
+
+        final ServletContextHandler context = new ServletContextHandler(ServletContextHandler.SESSIONS);
+        context.setContextPath("/");
+
+        final ServletContainer servletContainer = new ServletContainer(new JerseyAppConfig());
+        final ServletHolder servletHolder = new ServletHolder(servletContainer);
+        context.addServlet(servletHolder, "/*");
+        servletHolder.setInitOrder(0);
+        servletHolder.setInitParameter("" +
+                        "jersey.config.server.provider.classnames",
+                JerseyAppConfig.class.getCanonicalName());
+
+        this.jettyServer = new Server(JETTY_PORT);
+        this.jettyServer.setHandler(context);
+        this.jettyServer.start();
+
+    }
+
+    @SneakyThrows
+    public void stop() {
+        this.jettyServer.stop();
+    }
+
+    @SneakyThrows
+    private String getHostAddress() {
+        return InetAddress.getLocalHost().getHostAddress();
+    }
+
+    public String getBaseUrl() {
+        return "http://" + getHostAddress() + ":" + JETTY_PORT;
+    }
+}
+
--- a/src/systemTest/java/com/flightstats/hub/client/CallbackResourceClient.java
+++ b/src/systemTest/java/com/flightstats/hub/client/CallbackResourceClient.java
@@ -1,0 +1,11 @@
+package com.flightstats.hub.client;
+
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Path;
+
+public interface CallbackResourceClient {
+
+    @GET("/callback/{webhookName}")
+    Call<String> get(@Path("webhookName") String webhookName);
+}
--- a/src/systemTest/java/com/flightstats/hub/model/Channel.java
+++ b/src/systemTest/java/com/flightstats/hub/model/Channel.java
@@ -1,0 +1,26 @@
+package com.flightstats.hub.model;
+
+import lombok.Builder;
+import org.joda.time.DateTime;
+
+import java.util.Date;
+import java.util.Set;
+
+@Builder
+public class Channel {
+
+    private String name;
+    private String displayName;
+    private boolean keepForever;
+    private String owner;
+    private Date creationDate;
+    private long ttlDays;
+    private long maxItems;
+    private String description;
+    private Set<String> tags;
+    private String replicationSource;
+    private String storage;
+    private boolean protect;
+    private DateTime mutableTime;
+    private boolean allowZeroBytes;
+}
--- a/src/systemTest/java/com/flightstats/hub/model/Webhook.java
+++ b/src/systemTest/java/com/flightstats/hub/model/Webhook.java
@@ -1,0 +1,29 @@
+package com.flightstats.hub.model;
+
+import lombok.Builder;
+import lombok.experimental.Wither;
+
+@Builder
+public class Webhook {
+
+    private final String callbackUrl;
+    private final String channelUrl;
+
+    private final String name;
+    private final String batch;
+    private final boolean heartbeat;
+    private final boolean paused;
+    private final Integer ttlMinutes;
+    private final Integer maxWaitMinutes;
+    private final Integer callbackTimeoutSeconds;
+    private final boolean fastForwardable;
+    private final String tagUrl;
+    private final String managedByTag;
+    private final Integer maxAttempts;
+    private final String errorChannelUrl;
+
+    @Wither
+    private final String startItem;
+    @Wither
+    private final Integer parallelCalls;
+}
--- a/src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
+++ b/src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
@@ -1,0 +1,222 @@
+package com.flightstats.hub.testcase;
+
+import com.flightstats.hub.BaseTest;
+import com.flightstats.hub.callback.CallbackServer;
+import com.flightstats.hub.client.CallbackResourceClient;
+import com.flightstats.hub.client.ChannelItemResourceClient;
+import com.flightstats.hub.client.ChannelResourceClient;
+import com.flightstats.hub.client.WebhookResourceClient;
+import com.flightstats.hub.model.Channel;
+import com.flightstats.hub.model.ChannelItem;
+import com.flightstats.hub.model.Webhook;
+import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import retrofit2.Call;
+import retrofit2.Response;
+
+import javax.inject.Inject;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static com.flightstats.hub.model.ChannelContentStorageType.SINGLE;
+import static javax.ws.rs.core.Response.Status.CREATED;
+import static javax.ws.rs.core.Response.Status.OK;
+import static org.apache.commons.lang3.StringUtils.isBlank;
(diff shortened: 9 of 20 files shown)


src/systemTest/java/com/flightstats/hub/callback/CacheObject.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+@Slf4j
+public class CacheObject {
+
+    private static final Map<String, List<String>> cacheRequestObject = new ConcurrentHashMap<>();
```
chriskessel: Just making sure you want a global hashmap? Can an instance of this class be injected as a Singleton so you don't have a global?

src/systemTest/java/com/flightstats/hub/callback/CacheObject.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public void put(RequestObject requestObject) {
+        String webhookName = requestObject.getName();
+
+        synchronized (CacheObject.class) {
```
chriskessel: You've got a ConcurrentHashMap, but then you're using it in a synchronized block which sort of defeats the value of a ConcurrentHashMap.

src/systemTest/java/com/flightstats/hub/callback/CacheObject.java
```diff
@@ -0,0 +1,33 @@
...
+            if (cacheRequestObject.containsKey(webhookName)) {
```
chriskessel: I think this can be replaced with computeIfAbsent() and a concurrent safe list. In which case this entire method boils down to the following. Which is thread safe because both the hashmap and lists are concurrent safe versions.
```
public put(RequestObject requestObject) {
        String webhookName = requestObject.getName();
        cacheRequestObject.computeIfAbsent(webhookName, emptyList -> new CopyOnWriteArrayList()).addAll(requestObject.getUris());
}
```

An example:
```
        ConcurrentHashMap<String,List<String>> m = new ConcurrentHashMap<>();
        m.computeIfAbsent("foo", emptyList -> new CopyOnWriteArrayList<>()).addAll(Arrays.asList("a", "b"));
        m.computeIfAbsent("foo", emptyList -> new CopyOnWriteArrayList<>()).addAll(Arrays.asList("c", "d"));
        System.out.println(m.values());
```
Output is ```[[a, b, c, d]]```

src/systemTest/java/com/flightstats/hub/callback/CallbackResource.java
```diff
@@ -0,0 +1,47 @@
...
+
+        return Response.status(Response.Status.OK).build();
+    }
+
+    @GET
+    @Path("/{webhookname}")
+    @Produces(MediaType.TEXT_PLAIN)
+    @SneakyThrows
+    public Response get(@PathParam("webhookname") String webhookName) {
+        List<String> value = this.cacheObject.get(webhookName);
```
chriskessel: If CacheObject had a contains() method, this might read a little simpler as something like:
```
if (cacheObject.contains(webhookName)) {
    return Response.ok(cacheObject.get(webhookName.toString()).build();
} else {
    return Response.status(Response.Status.NOT_FOUND).build();
}
```

src/systemTest/java/com/flightstats/hub/model/Channel.java
```diff
@@ -0,0 +1,26 @@
+package com.flightstats.hub.model;
+
+import lombok.Builder;
+import org.joda.time.DateTime;
+
+import java.util.Date;
+import java.util.Set;
+
+@Builder
```
chriskessel: Maybe I'm missing something, but there's no ```@Data``` or ```@Value```, in which case it's all pure private values, so how does anyone get to the instance variables?

src/systemTest/java/com/flightstats/hub/model/Webhook.java
```diff
@@ -0,0 +1,25 @@
+package com.flightstats.hub.model;
+
+import lombok.Builder;
+
+@Builder
```
chriskessel: Same comment as before.

src/systemTest/java/com/flightstats/hub/callback/CacheObject.java
manjula-chikkanagappa: Thanks for reviewing the PR Chris. Appreciate it.

Yes I thought about singleton. It requires some setup with Guice, jersey and jetty. I am setting up jetty server to serve as a callback server. So did not want to get into the complicated stuff.
I really prefer spring framework any day which follows the mantra of convention over configuration and do not need to go thru all these hula hoops just to set up DI. But in this case I am simple did not want to introduce another framework in the current teach stack!

src/systemTest/java/com/flightstats/hub/callback/CacheObject.java
manjula-chikkanagappa: synchronized is for code checking the map. That is not synchronized.

src/systemTest/java/com/flightstats/hub/callback/CacheObject.java
chriskessel: Heh, different strokes I guess. I've used Spring and thought Guice was much simpler.

src/systemTest/java/com/flightstats/hub/callback/CacheObject.java
manjula-chikkanagappa: That looks lot nicer.  In this case, there will put multiple times for the same key. Similar to Multimap.

src/systemTest/java/com/flightstats/hub/model/Channel.java
manjula-chikkanagappa: For now these objects are not being used for reading. Only for creating for POST request body. This will evolve over time :)

src/systemTest/java/com/flightstats/hub/callback/CallbackResource.java
manjula-chikkanagappa: I like this better. I will change this.

src/systemTest/java/com/flightstats/hub/model/Channel.java
chriskessel: Oh! It's purely for serialization? That makes sense.

src/systemTest/java/com/flightstats/hub/callback/CacheObject.java
manjula-chikkanagappa: Could you please share the repo url?

chriskessel: https://bitbucket.flightglobal.io/projects/PDX-DS/repos/flex/browse

The different directories are different products, but there's a bunch of code shared between them. We use Dropwizard, which we have a few lines in an initalize() method to plug in Guice capability.

After that, it's just all using ```@Inject``` and, for things that aren't trivial to construct, ```@Provides``` methods. 

You can see one of our modules here. None of that may be terribly relevant though to you since the Hub has it's own framework of some sort.
https://bitbucket.flightglobal.io/projects/PDX-DS/repos/flex/browse/connections/src/main/java/com/flightstats/connections/ConnectionsMain.java
https://bitbucket.flightglobal.io/projects/PDX-DS/repos/flex/browse/connections/src/main/java/com/flightstats/connections/GuiceModule.java

manjula-chikkanagappa: Thanks Chris. I let go of my laziness and implemented DI. Looks much better now :)
Thank you!

src/systemTest/java/com/flightstats/hub/client/CallbackResourceClient.java
```diff
@@ -0,0 +1,11 @@
+package com.flightstats.hub.client;
+
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Url;
+
+public interface CallbackResourceClient {
+
+    @GET
+    Call<String> get(@Url String hostUrl);
```
lkemmerer: Is there a reason this is so generic compared to the hub clients? It seems like creating a retrofit client with the callback server base uri would let you make this more explicit, with a path of `\{webhookName}`. That would make the test below read a little more cleanly, and then you could specify that it's going to be a `List<String>` which you wouldn't have to parse manually...

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+    private List<String> channelItemsPosted;
+
+    @Before
+    public void setup() {
+
+        this.channelItemResourceClient = getHttpClient(ChannelItemResourceClient.class);
+        this.channelResourceClient = getHttpClient(ChannelResourceClient.class);
+        this.webhookResourceClient = getHttpClient(WebhookResourceClient.class);
+        this.callbackResourceClient = getHttpClient(CallbackResourceClient.class);
```
lkemmerer: Would it make sense to use a different retrofit client builder for the callback client, since it seems like it'll have a different base url? Trips created a whole Retrofit client builder to make that a little easier (it's in bitbucket under PDX-TRIPS -> trips-common -> RetrofitClientBuilder.java), though I think it's for retrofit v1, so it'd need tweaking.

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
```diff
@@ -0,0 +1,198 @@
...
+        createChannel();
+        List<String> channelItems = addItemsToChannel(data);
+
+        addWebhook(buildWebhook(channelItems.get(4)));
+        verifyWebhookCallback(channelItems.subList(5, channelItems.size()));
+    }
+
+    @SneakyThrows
+    private void logWebhookCallbackError() {
+        log.info("Webhook name {} ", webhookName);
```
lkemmerer: Is this meant to test that errors are logged when a callback returns something other than a 2xx? If so, I'd probably create another callback endpoint that always returns a 4xx or 5xx and create a webhook with that as a callbackUrl That way when you add an item to the channel, you can confirm that the item uri shows up in the webhook errors list.  I think it'd mean you'll need to create a WebhookErrors model, though.

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+
+    private void verifyWebhookCallback(List<String> channelItems) {
+        Call<String> call = callbackResourceClient.get(callbackServer.getUrl() + "/" + this.webhookName);
+
+        try {
+            await().atMost(90, TimeUnit.SECONDS).until(() -> {
+                Response<String> response = call.clone().execute();
+                channelItemsPosted = parseResponse(response.body());
```
lkemmerer: It looks like `channelItemsPosted` can be a variable defined in this method rather than a member of the class.

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
```diff
@@ -0,0 +1,198 @@
...
+                return response.code() == OK.getStatusCode()
+                        && channelItemsPosted.size() == channelItems.size();
+            });
+        } catch (Exception e) {
+            log.error("Problem verifying webhook callbacks. {} ", e.getMessage());
+            logWebhookCallbackError();
+        }
+
+        Collections.sort(channelItems);
```
lkemmerer: Why does this method sort the items? By doing that, this doesn't verify that the callback receives items in the list they're sent.

src/systemTest/java/com/flightstats/hub/callback/model/RequestObject.java
```diff
@@ -0,0 +1,13 @@
+package com.flightstats.hub.callback.model;
+
+import lombok.Data;
+
+import java.util.List;
+
+@Data
+public class RequestObject {
+    //webhook name
+    private String name;
```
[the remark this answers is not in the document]
chriskessel: +1, though it probably means an annotation to make it serialize "name" as the field since I'm guessing that "name" is part of the spec'd callback api?

src/systemTest/java/com/flightstats/hub/callback/CallbackResource.java
```diff
@@ -0,0 +1,54 @@
...
+    @Consumes(MediaType.APPLICATION_JSON)
+    @SneakyThrows
+    public Response create(RequestObject requestObject) {
+        log.info("Callback request received {} ", requestObject);
+        this.cacheObject.put(requestObject);
+        return Response.status(Response.Status.OK).build();
+    }
+
+    @GET
+    @Path("/{webhookname}")
```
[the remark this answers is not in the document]
chriskessel: I'm not sure if there's an official REST policy, but yea, I make the names the same as I'd name them in the Java code. Probably "webhookName" in this case?

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    private Webhook buildWebhook(String startItem) {
+        Webhook.WebhookBuilder builder = Webhook.builder()
+                .channelUrl(retrofit.baseUrl() + "channel/" + channelName)
+                .callbackUrl(callbackServer.getUrl())
+                .parallelCalls(2)
+                .batch("single");
+
+        if (!isBlank(startItem)) {
```
chriskessel: Personal preference, but I'd just inline this with a tertiary and remove the if clause. Gets rid of the intermediate builder variable.
```
return Webhook.builder()
  .channelUrl(retrofit.baseUrl() + "channel/" + channelName)
  .callbackUrl(callbackServer.getUrl()).callbackUrl(callbackServer.getUrl())
  .parallelCalls(2)
  .batch("single") // Probably for a future day, but seems like this should be an enum
  .startItem(isBlank() ? null : startItem)
.build()
```

src/systemTest/java/com/flightstats/hub/callback/model/RequestObject.java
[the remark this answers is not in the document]
lkemmerer: I think the callback API is whatever we define it as, but I might be mistaken.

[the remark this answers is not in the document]
chriskessel: If this is something that your customers see, then the current "name" is a contract you wouldn't want to break. Someone creating a request is going to be creating it as "name".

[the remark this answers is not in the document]
manjula-chikkanagappa: Yes. the it comes in as "name" in the POST request body.

src/systemTest/java/com/flightstats/hub/callback/model/RequestObject.java
```diff
@@ -0,0 +1,13 @@
...
+    private List<String> uris;
```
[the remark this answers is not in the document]
manjula-chikkanagappa: Yes. the it comes in as "uris" in the POST request body.

src/systemTest/java/com/flightstats/hub/model/Channel.java
manjula-chikkanagappa: I am not sure how we will be using this model. whether we really need it or not. It depends on the testcases. I would follow pirnciple of YAGNI here :)

build.gradle
```diff
@@ -10,6 +10,22 @@ repositories {
     mavenCentral()
 }
 
+sourceSets {
+    systemTest {
+        java {
+            compileClasspath += main.output + test.output
+            runtimeClasspath += main.output + test.output
+            srcDir file('src/integration-test/java')
```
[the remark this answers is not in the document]
manjula-chikkanagappa: Good catch laurie!.

src/systemTest/java/com/flightstats/hub/client/CallbackResourceClient.java
manjula-chikkanagappa: I need to know the callback server url to create a retrofit client. callback instance gets injected here .
Also response parsing limitation is on retrofit side. Idid not see any special converters for java data structure  type interestingly.
        return new Retrofit.Builder()
                    .baseUrl(hubBaseUrl)
                    .addConverterFactory(ScalarsConverterFactory.create())
                    .addConverterFactory(GsonConverterFactory.create())
                    .client(new OkHttpClient.Builder().build())
                    .build();

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
manjula-chikkanagappa: lists compare works only when 2 lists are sorted.

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
manjula-chikkanagappa: i know :(
Lambda function ccepts local variable only when it is final!

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
manjula-chikkanagappa: this is not test for errors endpoint. when exception happens I am trying to log errors. After test chanel, webhook is all cleaned up and we do not have data to verify.

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
chriskessel: Yea, you can't change a variable that's defined outside the lambda. You could probably do this though:
```channelItemsPosted.addAll(parseResponse(response.body());```
or declare channelItems via AtomicReference:
```channelItemsPosted = new AtomicReference();```
The set that in your lambda:
```channelItemsPosted.set(parseResponse(response.body());```

File: src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java [hunk 17]
```diff
@@ -0,0 +1,198 @@
...
+    @Before
+    public void setup() {
+
+        this.channelItemResourceClient = getHttpClient(ChannelItemResourceClient.class);
+        this.channelResourceClient = getHttpClient(ChannelResourceClient.class);
+        this.webhookResourceClient = getHttpClient(WebhookResourceClient.class);
+        this.callbackResourceClient = getHttpClient(CallbackResourceClient.class);
+        this.callbackServer = injector.getInstance(CallbackServer.class);
+
+        this.callbackServer.start();
```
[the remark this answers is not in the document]
manjula-chikkanagappa: Junit's beforeclass and after class methods needs to be static. This in turn requires all variables used in the method to be static. Ideally callback server should be started once before all tests

src/systemTest/java/com/flightstats/hub/callback/model/RequestObject.java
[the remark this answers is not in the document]
lkemmerer: Oh, I see. For some reason I thought this was something that was internal to our callback server and not what the webhook returned to the callback. Duh.

src/systemTest/java/com/flightstats/hub/callback/model/RequestObject.java
```diff
@@ -0,0 +1,13 @@
+package com.flightstats.hub.callback.model;
+
+import lombok.Data;
+
+import java.util.List;
+
+@Data
+public class RequestObject {
```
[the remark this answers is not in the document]
lkemmerer: If this is a model that the hub exposes as part of the webhook callback rather than it being an internal part of the callback server we have, I'd move it up to hub/model instead of hub/callback/model.

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+        assertEquals(CREATED.getStatusCode(), response.code());
+        assertNotNull(response);
+        assertNotNull(response.body());
+
+        return response.body().get_links().getSelf().getHref();
+    }
+
+    private Webhook buildWebhook(String startItem) {
+        Webhook.WebhookBuilder builder = Webhook.builder()
```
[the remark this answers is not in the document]
lkemmerer: Er, `@Wither`. Gah. Words are hard.

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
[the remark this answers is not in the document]
lkemmerer: If the intent is for the callback server to be started once for this test file, you're right that the server will need to be static so you can start it in a `@BeforeClass`.

src/systemTest/java/com/flightstats/hub/client/CallbackResourceClient.java
lkemmerer: I'm not sure I follow. Would it be possible to start the server and then instantiate the retrofit client, since at that point you'd have the url?

lkemmerer: By special converter for Java data structures, do you mean that it can't parse a `List<String>` but it can parse a `Webhook`?

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
lkemmerer: Oh! I didn't read closely enough. Sorry about that.

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
lkemmerer: That is odd. I think one of the expectations we have for webhooks is that they'll arrive in order unless there's a failure and retrying is involved. Does the test pass if you sort `channelItems` but not `channelItemsPosted`?

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
```diff
@@ -62,25 +64,49 @@ public void setup() {
...
 
-        addWebhook(buildWebhook(channelItems.get(4)));
-        verifyWebhookCallback(channelItems.subList(5, channelItems.size()));
+        addWebhook(buildWebhook(channelItems.get(4), 2));
+        final List<String> channelItemsExpected = channelItems.subList(5, channelItems.size());
+        final List<String> channelItemsPosted = getWebhookCallbackItems(channelItemsExpected.size());
+
+        Collections.sort(channelItemsExpected);
+        Collections.sort(channelItemsPosted);
+        assertEquals(channelItemsExpected, channelItemsPosted);
```
lkemmerer: I think you can use hamcrest's `assertThat(channelItemsPosted, containsInAnyOrder(channelItemsExpected.toArray()))` here to make this a little more readable.

manjula-chikkanagappa: I know. I liked that too. But there is a conflict while adding hamcrest dependency currently. it needs to be excluded from one of the existing dependencies. So I am sticking this simple solution now.

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
```diff
@@ -172,11 +192,11 @@ private String addItemToChannel(Object data) {
...
+    private Webhook buildWebhook(String startItem, int parallelCalls) {
         return Webhook.builder()
-                .channelUrl(retrofit.baseUrl() + "channel/" + channelName)
-                .callbackUrl(callbackServer.getUrl())
-                .parallelCalls(2)
+                .channelUrl(getHubClientBaseUrl() + "channel/" + channelName)
+                .callbackUrl(getCallbackClientBaseUrl() + "callback/")
+                .parallelCalls(parallelCalls)
                 .batch("single")
                 .startItem(isBlank(startItem) ? null : startItem)
```
lkemmerer: Did you decide to skip adding a `@Wither` annotation because it makes the API of the `Webhook` class ugly? I would rather explicitly override the start item in the test because going down the path of adding params gets hard to read. Having a method that sets some defaults and relies on the calling test to specify the fields that are important for each test.

src/systemTest/java/com/flightstats/hub/callback/CallbackCache.java
```diff
@@ -0,0 +1,37 @@
...
+@Slf4j
+public class CallbackCache {
+
+    private final Map<String, List<String>> cacheRequestObject = new ConcurrentHashMap<>();
+
+    public void put(WebhookCallbackRequest webhookCallbackRequest) {
+        String webhookName = webhookCallbackRequest.getName();
+
+        synchronized (this) {
+            if (cacheRequestObject.containsKey(webhookName)) {
```
lkemmerer: Could you use the public `containsKey` and `get` methods below here?

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
lkemmerer: Ugh, that is sad. Sounds good.

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
manjula-chikkanagappa: Added @Wither annotation

README.md
```diff
@@ -34,3 +34,24 @@ The hub is a long term data store.
...
 
 The above command compiles the code, runs the unit tests and integration tests.~~~~
+
+
+### How to run System tests:
+
+From Local:
+
+1) Setup hub:
+    docker run -p 80:80 flightstats/hub:latest
```
Paul-Hess: minor nitpick, but I think this should be something like ... flightstats/hub:local. The image created by the docker shell script is name spaced local because we used to call everything 'latest' which caused some confusing conflicts if you pulled the image from our docker repo, which is more rightly called latest

src/systemTest/java/com/flightstats/hub/callback/CallbackCache.java
```diff
@@ -0,0 +1,37 @@
...
+    private final Map<String, List<String>> cacheRequestObject = new ConcurrentHashMap<>();
+
+    public void put(WebhookCallbackRequest webhookCallbackRequest) {
+        String webhookName = webhookCallbackRequest.getName();
+
+        synchronized (this) {
+            if (contains(webhookName)) {
+                get(webhookName).addAll(webhookCallbackRequest.getUris());
+            } else {
+                cacheRequestObject.put(
```
Paul-Hess: I think we will have a need in future for this cache to handle multiple sets of values under the same webhook name, I adding some kind of further unique id to the key side of the pair would be necessary in the future if not now. Not suggesting it is in the scope of this PR necessarily, but worth a mention.

src/systemTest/java/com/flightstats/hub/callback/CallbackServer.java
```diff
@@ -0,0 +1,50 @@
...
+public class CallbackServer {
+
+    private static final int JETTY_PORT = 8090;
+    private Server jettyServer;
+
+    @SneakyThrows
+    public void start() {
+
+        final ServletContextHandler context = new ServletContextHandler(ServletContextHandler.SESSIONS);
+        context.setContextPath("/");
```
Paul-Hess: If we want to end up running these tests concurrently, or even if we don't but less often, using a base path for all callbacks will result at some point in leakage between tests unfortunately. I'd like to see this at some point having a unique context path for each test, which could probably be injected at creation of the server. Another thing that doesn't have to be in this PR necessarily but worth mentioning

src/systemTest/java/com/flightstats/hub/testcase/WebhookLifecycleTest.java
Paul-Hess: I am not sure that we should be sorting these, generally our webhook assertions should confirm that the order that the callbacks happened is correct and sorting I believe will tamper with that

Paul-Hess: I was able to use Hamcrest (on another branch), I had add it's import before junit in the gradle file, and that worked.
Not sure if that is the same conflict you are seeing now. Based on my comment above I'd prefer a different matcher than containsInAnyOrder though :)

PR #1147, opened 2019-03-08
PDXDDT-683 Replace TestNG library with Junit library

manjula-chikkanagappa:
Removed TestNG library references. Now we have only Junit testing framework in our hub code.

### DIFF
--- a/src/test/java/com/flightstats/hub/webhook/TagWebhookTest.java
+++ b/src/test/java/com/flightstats/hub/webhook/TagWebhookTest.java
@@ -3,8 +3,8 @@ package com.flightstats.hub.webhook;
 import com.flightstats.hub.model.ChannelConfig;
 import com.flightstats.hub.test.Integration;
 import org.apache.commons.lang3.StringUtils;
-import org.testng.annotations.BeforeSuite;
-import org.testng.annotations.Test;
+import org.junit.Before;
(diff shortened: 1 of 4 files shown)


src/test/java/com/flightstats/hub/webhook/TagWebhookTest.java
```diff
@@ -54,7 +54,7 @@ private ChannelConfig createChannelConfig(String name, String tag) {
                 .build();
     }
 
-    @BeforeSuite
+    @Before
```
lkemmerer: If I remember correctly, junit's `@Before` runs before each test. `@BeforeClass` should be the equivalent of `@BeforeSuite`.  Junit5 is much more intuitive, with `@BeforeEach` and `@BeforeAll`.

manjula-chikkanagappa: you are right Laurie. There is only one test in this class. So it does not matter. Junit's Before class requires Beforeclass methods to be static.

lkemmerer: Ah, I didn't realize there was only one test... Honestly I don't think I would have added a separate setup method for a single test. I know you didn't write this; just thinking out loud.

PR #1149, opened 2019-03-12
Metric filter return all

Paul-Hess:


src/main/java/com/flightstats/hub/metrics/StatsdReporter.java
```diff
@@ -23,18 +23,26 @@ public StatsdReporter(
         this.dataDogHandler = dataDogHandler;
     }
 
+    // unused but useful in future. currently default reporting to both clients for all metrics
```
lkemmerer: I'm not sure I'd bother with the comment. Even the lazy can look up method usages and see that there are none. :) And if there are eventually some, this comment won't be true anymore and someone will need to remember to delete it. If you wanted to put an explanation somewhere, I'd probably put it in the merge message for the PR about what we're doing and why we're doing it.

PR #1151, opened 2019-04-05
Ensure we enqueue S3 writes after writing to the spoke.

dkleckler:


src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.mockito.runners.MockitoJUnitRunner;
+
+import java.util.Optional;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ClusterContentServiceTest {
+    @Mock private ClusterContentService ccs;
```
lkemmerer: I didn't know about this annotation. That's useful!

src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
```diff
@@ -0,0 +1,96 @@
...
+    @Mock private ChannelService channelSvc;
+    @Mock private ContentDao mockSpokeWriteDao, mockSpokeReadDao;
```
lkemmerer: I don't know how everybody else feels about multiple member declarations on one line. I think I prefer separate lines even if the ones you've grouped are related; it seems more consistent and therefore less surprising. Others should weigh in here, though, since we don't have an explicit convention related to this.

src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Before
+    public void initClusterContentService() {
+        channelName = "/testChannel";
+        when(channelSvc.getCachedChannelConfig(channelName)).thenReturn(Optional.of(channelConfig));
+        ccs = new ClusterContentService(channelSvc, mockSpokeWriteDao, mockSpokeReadDao, mockS3SingleDao, mockS3LargeDao, mockS3BatchDao, s3WriteQueue, lastContentPath, hubUtils);
+    }
+
+    @SneakyThrows
+    private void initSimpleContent(boolean isLarge) {
```
lkemmerer: I usually put private methods below public ones, though I do get that it's helpful to know what helper methods are doing before they're used.

src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
dkleckler: Yeah, saves a little boilerplate on the mocking.

src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
```diff
@@ -0,0 +1,96 @@
...
+        byte[] data = "SimpleTest".getBytes();
+        contentKey = new ContentKey(TimeUtil.now(), "someHash");
+        content = Content.builder()
+                .withContentType("text/plain")
+                .withContentKey(contentKey)
+                .withLarge(isLarge)
+                .withData(data).build();
+        when(mockSpokeWriteDao.insert(anyString(), any(Content.class))).thenReturn(contentKey);
+        when(channelConfig.isBatch()).thenReturn(false);
```
lkemmerer: I know this DRYs a couple tests up a bit, but I think it's unrelated to setting up the content key.  Also, since it's a "given" that ties directly to the assertions, I'd usually prefer to make it explicit in each test so that there's symmetry across all the use cases.

src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
dkleckler: Yeah, I usually prefer individual lines too.  In this case it felt like it would help clarify similarities amongst a bunch of similarly-typed vars.  I'm not invested though; whichever you all prefer works for me.

src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
dkleckler: Sure thing.  I'll move this down.

src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
dkleckler: Yeah, that's fair.  Moving.

src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
lkemmerer: If you'd usually leave it as two, being consistent probably trumps the grouping, imo. The names are similar enough that it shouldn't be too bad to associate them. 🤷‍♂️

dkleckler: s/this case/cases like this, probably for me...but not everyone might agree with the grouping...it might be more aligned with how I think through this class.  🤷‍♂️ as well.  I'll split them out.

crozierm: The convention in the Portland office is to have one per line, though I'm not sure it's written down anywhere. And I think it's generally looked down upon in the Java community.

src/main/java/com/flightstats/hub/dao/aws/ClusterContentService.java
```diff
@@ -63,37 +64,41 @@
...
+    public ClusterContentService(
+                            ChannelService channelService,
+                            @Named(ContentDao.WRITE_CACHE) ContentDao spokeWriteContentDao,
+                            @Named(ContentDao.READ_CACHE) ContentDao spokeReadContentDao,
+                            @Named(ContentDao.SINGLE_LONG_TERM) ContentDao s3SingleContentDao,
+                            @Named(ContentDao.LARGE_PAYLOAD) ContentDao s3LargePayloadContentDao,
+                            @Named(ContentDao.BATCH_LONG_TERM) ContentDao s3BatchContentDao,
+                            S3WriteQueue s3WriteQueue,
+                            LastContentPath lastContentPath,
+                            HubUtils hubUtils) {
```
manjula-chikkanagappa: Nice. Constructor based injection is more readable and definitely a better choice since it throws exception sooner if one of the dependencies is not available.

src/main/java/com/flightstats/hub/model/LargeContent.java
```diff
@@ -3,17 +3,16 @@
...
 
 import java.io.IOException;
 
+@Slf4j
 public class LargeContent {
 
     static final String CONTENT_TYPE = "application/hub";
-    private static final Logger logger = LoggerFactory.getLogger(LargeContent.class);
-    private static final ObjectMapper mapper = HubProvider.getInstance(ObjectMapper.class);
+    private static final ObjectMapper mapper = HubBindings.objectMapper();
```
manjula-chikkanagappa: This should be injected to maintain consistency across the app.

dkleckler: Yeah, my preference was to switch this to an injected singleton, but guicification was starting to snowball.  I can go do that though.
Do we have a general route we're going here?  I see lots of cases where the code pulls an instance manually from the Injector through HubProvider.  Not crazy about that pattern.

lkemmerer: We're trying to deprecate HubProvider, so your instinct not to use it is appreciated. Lemme take a look at the code and think a little before I answer the question about whether we have a general route. In the meantime, maybe someone who's feeling more coherent has an answer more readily available.

Honestly, if it was starting to snowball, I'd be cool with splitting that refactoring out to a different PR. Makes it easier to revert (especially if we squash-merge our PRs).

dkleckler: It's not that bad.  I agree it's definitely better this way.  I'll toss it in.  CCS is the only consumer, so it doesn't reach that far.  Will push again to DI tonight though since there's been a bit of churn.

lkemmerer: Eww, all the methods in here are static. I'd probably make them instance methods and start injecting the ObjectMapper here and a LargeContent dependency in `ClusterContentService`...which I think is the only place it's used. Does that seem like it'd work or is that oversimplified?

src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
```diff
@@ -0,0 +1,105 @@
+package com.flightstats.hub.dao.aws;
+
+import com.flightstats.hub.cluster.LastContentPath;
+import com.flightstats.hub.dao.ChannelService;
+import com.flightstats.hub.dao.ContentDao;
+import com.flightstats.hub.model.*;
```
Paul-Hess: As a rule we prefer to expand all splat imports

src/main/java/com/flightstats/hub/model/LargeContent.java
Paul-Hess: Plus one to doing this in another PR, seems a little out of scope

dkleckler: Ah...sorry...just sent it up.

src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
dkleckler: Yeah, fixed IJ's min-5 for splat rule.

PR #1153, opened 2019-04-09
Ensure writes to UDP port happen after we start listening.

dkleckler:


src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java
```diff
@@ -42,15 +43,20 @@ private static StatsdReporter provideStatsDHandlers() {
         return provider.get();
     }
 
+    private static void runAfterSocketInit(Runnable runnable) {
+        Executors.newScheduledThreadPool(1).schedule(runnable, 100, TimeUnit.MILLISECONDS);
```
chriskessel: I'm surprised this works. The returned Executor is local to the the function scope and I'd think it'd get garbage collected and thus stop rescheduling. Maybe sufficiently reliable in a unit test though.

crozierm:
I'd be tempted to remove the static cancer from this class. I'll admit I didn't look at it very closely though.

src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java
dkleckler: ScheduledThreadPoolExecutor has an onShutdown() hook that ensures delayed tasks execute even after shutdown (by default).  Periodic tasks will be canceled though.

Paul-Hess:
@BeforeClass must be static, I find it useful, particularly here

dkleckler:
Laurie killed off all the timed waits here, which should result in an optimally fast test execution.  Closing this one to PR over from her branch.

PR #1154, opened 2019-04-09
Add some concurrency-related safeties for StatsdReporter test

lkemmerer:
After Damon's changes, I was still getting consistent failures here. I threw some more concurrency-related stuff in and removed `Thread.sleep`s. It seems more stable now.

In the process of debugging, I ended up inlining the BeforeClass method, which likely decreases debuggability, but I found it difficult with all the `CompletableFuture`s already. :-/

### DIFF
--- a/src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java
+++ b/src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java
@@ -6,7 +6,6 @@ import com.flightstats.hub.dao.Dao;
 import com.flightstats.hub.model.ChannelConfig;
 import com.flightstats.hub.util.IntegrationUdpServer;
 import com.flightstats.hub.webhook.Webhook;
-import lombok.extern.slf4j.Slf4j;
 import org.junit.Test;
 
 import java.util.Map;
@@ -17,15 +16,14 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 import static org.mockito.Mockito.mock;
 
 public class StatsdReporterIntegrationTest {
-    private String[] tags = { "tag1", "tag2" };
-    private MetricsConfig metricsConfig = MetricsConfig.builder()
+    private final String[] tags = { "tag1", "tag2" };
+    private final MetricsConfig metricsConfig = MetricsConfig.builder()
             .hostTag("test_host")
             .statsdPort(8123)
             .dogstatsdPort(8122)
@@ -36,25 +34,6 @@ public class StatsdReporterIntegrationTest {
 
     private final IntegrationUdpServer udpServer = provideNewServer(metricsConfig.getStatsdPort());
     private final IntegrationUdpServer udpServerDD = provideNewServer(metricsConfig.getDogstatsdPort());
-
-    private IntegrationUdpServer provideNewServer(int port) {
-        return IntegrationUdpServer.builder()
-                .listening(true)
-                .port(port)
-                .startupCountDownLatch(startupCountDownLatch)
-                .executorService(executorService)
-                .build();
-    }
-
-    @SuppressWarnings("unchecked")
-    private StatsdReporter provideStatsDHandlers() {
-        Dao<ChannelConfig> channelConfigDao = (Dao<ChannelConfig>) mock(CachedLowerCaseDao.class);
-        Dao<Webhook> webhookDao =  (Dao<Webhook>) mock(CachedDao.class);
-        StatsDFilter statsDFilter = new StatsDFilter(metricsConfig, channelConfigDao, webhookDao);
-        statsDFilter.setOperatingClients();
-        StatsDReporterProvider provider = new StatsDReporterProvider(statsDFilter, metricsConfig);
-        return provider.get();
-    }
 
     @Test
     public void StatsDHandlersCount_metricShape() throws InterruptedException, ExecutionException, TimeoutException {
@@ -69,6 +48,15 @@ public class StatsdReporterIntegrationTest {
         assertEquals("hub.countTest:1|c|#tag2,tag1", resultsDogStatsd.get("hub.countTest"));
     }
 
+    private IntegrationUdpServer provideNewServer(int port) {
+        return IntegrationUdpServer.builder()
+                .listening(true)
+                .port(port)
+                .startupCountDownLatch(startupCountDownLatch)
+                .executorService(executorService)
+                .build();
+    }
+
     private CompletableFuture<String> getMetricsWriterFuture() {
         return CompletableFuture.supplyAsync(() -> {
             try {
@@ -80,6 +68,17 @@ public class StatsdReporterIntegrationTest {
             return "done";
         }, executorService);
     }
+
+    @SuppressWarnings("unchecked")
+    private StatsdReporter provideStatsDHandlers() {
+        Dao<ChannelConfig> channelConfigDao = (Dao<ChannelConfig>) mock(CachedLowerCaseDao.class);
+        Dao<Webhook> webhookDao =  (Dao<Webhook>) mock(CachedDao.class);
+        StatsDFilter statsDFilter = new StatsDFilter(metricsConfig, channelConfigDao, webhookDao);
+        statsDFilter.setOperatingClients();
+        StatsDReporterProvider provider = new StatsDReporterProvider(statsDFilter, metricsConfig);
+        return provider.get();
+    }
+
 
     private void writeMetrics() {
         StatsdReporter handlers = provideStatsDHandlers();
--- a/src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java
+++ b/src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java
@@ -6,6 +6,7 @@ import com.flightstats.hub.dao.Dao;
 import com.flightstats.hub.model.ChannelConfig;
 import com.flightstats.hub.util.IntegrationUdpServer;
 import com.flightstats.hub.webhook.Webhook;
+import lombok.SneakyThrows;
 import org.junit.Test;
 
 import java.util.Map;
@@ -35,13 +36,14 @@ public class StatsdReporterIntegrationTest {
     private final IntegrationUdpServer udpServer = provideNewServer(metricsConfig.getStatsdPort());
     private final IntegrationUdpServer udpServerDD = provideNewServer(metricsConfig.getDogstatsdPort());
 
+    @SneakyThrows
     @Test
-    public void StatsDHandlersCount_metricShape() throws InterruptedException, ExecutionException, TimeoutException {
+    public void StatsDHandlersCount_metricShape() {
         CompletableFuture.allOf(
                 getMetricsWriterFuture(),
                 udpServer.getServerFuture(startupCountDownLatch, executorService),
                 udpServerDD.getServerFuture(startupCountDownLatch, executorService)
-        ).get(15000, TimeUnit.MILLISECONDS);
+        ).get(5000, TimeUnit.MILLISECONDS);
 
         Map<String, String> resultsStatsd = udpServer.getResult();
         assertEquals("hub.countTest:1|c|#tag2,tag1", resultsStatsd.get("hub.countTest"));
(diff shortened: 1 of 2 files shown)


src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
```diff
@@ -1,74 +1,71 @@
...
-        }
-//        logger.info(":::::::::, {}", store);
-        return store;
-    });
+    public CompletableFuture<Map<String, String>> getServerFuture() {
+        return  CompletableFuture.supplyAsync(() -> {
+            try {
+                DatagramSocket serverSocket = new DatagramSocket(port);
+                while (listening) {
+                    byte[] data = new byte[70];
```
chriskessel: If you know what the magic number 70 means, it's worth extracting a constant while you're here.

src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
```diff
@@ -1,74 +1,71 @@
...
-                    .lines()
-                    .collect(Collectors.toList())
-                    .get(0)
-                    .trim();
-            if (result.contains("closeSocket")) {
+                    String result = listen(receivePacket);
+                    addValueToStore(result);
+                }
+            } catch (IOException e) {
                 listening = false;
```
chriskessel: listening only goes to false on IOException, not other exceptions?

src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
```diff
@@ -1,74 +1,71 @@
...
 
+@Slf4j
 @Builder
 public class IntegrationUdpServer {
-    private final static Logger logger = LoggerFactory.getLogger(IntegrationUdpServer.class);
-    private long timeoutMillis;
-    private final List<String> queue = new ArrayList<>();
     private int port;
     private boolean listening;
+    private CountDownLatch startupCountDownLatch;
```
chriskessel: All of these instance variables, are they used across threads? None are volatile or wrapped in AtomicReference so there'd be potential race conditions or lack of cross-thread visibility guarantees.

src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
```diff
@@ -1,74 +1,71 @@
...
+    }
+
+    private String listen(DatagramPacket receivePacket) throws IOException {
+        InputStream inputStream = new ByteArrayInputStream(receivePacket.getData());
+        InputStreamReader streamReader =  new InputStreamReader(inputStream, StandardCharsets.UTF_8);
+        BufferedReader reader = new BufferedReader(streamReader);
+        String result = reader
+                .lines()
+                .collect(Collectors.toList())
+                .get(0)
```
chriskessel: Can you use Stream.findFirst()? No point in reading all the lines() if you just need the first one. Unless a full read is required to clear the socket?

src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
```diff
@@ -1,74 +1,71 @@
...
+                .trim();
+
+        if (result.contains("closeSocket")) {
+            listening = false;
         }
+
+        reader.close();
```
chriskessel: You could use try-with-resources, right?

lkemmerer: I had that in one of my fix attempts and forgot to add it back in. 👍

src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
lkemmerer: Would it be acceptable to turn these into parameters on the method that uses them and make them final there instead? The test that uses this creates them as final members... I don't have a great grasp of final vs volatile when it comes to visibility and mutations.

src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
lkemmerer: I'm afraid I don't know. @Paul-Hess?

Paul-Hess: At some point it was the number of bytes in the fairly static string sent to the server

src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
chriskessel: Either final or volatile will work. Final variables are guaranteed to be visible once construction is finished, since they're write once the JVM notifies all other threads of the final value. Volatile would also work for visibility because that tells the JVM to always check the value of the variable and never rely on what the thread has cached.

Final is going to be more performant if that matters here. It's also simpler to reason about since something that never changes can't have race conditions. I'd suggest making "listening" ```final AtomicBoolean```.

If you do any multi-thread work, it's definitely worth reading the chapter about visibility in Java Concurrency in Practice.

chriskessel: Edit: Oh, you asked about final method parameters? Method parameters don't have concurrency concerns, other than the things you pass could themselves be used in other threads.

```final``` is pretty pointless for method parameters or variables in a method. It just prevents you from assigning a new value to that parameter...which is generally bad form anyway (not always, but usually). ```final``` for method parameters doesn't have any impact for concurrency. I don't think ```volatile``` even applies for method parameters, I believe that's only for instance variables.

File: src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java [hunk 6]
```diff
@@ -43,26 +56,43 @@ private static StatsdReporter provideStatsDHandlers() {
...
     }
 
-    private static void runAfterSocketInit(Runnable runnable) {
-        Executors.newScheduledThreadPool(1).schedule(runnable, 100, TimeUnit.MILLISECONDS);
-    }
-
-
-    @BeforeClass
-    public static void startMockStatsDServer() {
+    public void startMockStatsDServer() throws InterruptedException, ExecutionException, TimeoutException {
```
manjula-chikkanagappa: InterruptedException is already caught in catch block. We don't need this throw. We should just use lombok's sneakyThrows annotation

src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java
```diff
@@ -43,26 +56,43 @@ private static StatsdReporter provideStatsDHandlers() {
...
         StatsdReporter handlers = provideStatsDHandlers();
-        IntegrationUdpServer udpServer = provideNewServer(metricsConfig.getStatsdPort());
-        IntegrationUdpServer udpServerDD = provideNewServer(metricsConfig.getDogstatsdPort());
-        runAfterSocketInit(() -> {
+        ExecutorService executorService = Executors.newFixedThreadPool(3);
+        IntegrationUdpServer udpServer = provideNewServer(metricsConfig.getStatsdPort(), startupCountDownLatch, canListen, executorService);
+        IntegrationUdpServer udpServerDD = provideNewServer(metricsConfig.getDogstatsdPort(), startupCountDownLatch, canListen, executorService);
+        CompletableFuture<String> sender = CompletableFuture.supplyAsync(() -> {
+            try {
+                startupCountDownLatch.await(15000, TimeUnit.MILLISECONDS);
```
manjula-chikkanagappa: 15 sec That's lot of time.  What exactly is this doing?  Even jetty server comes up in few seconds :)

src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java
```diff
@@ -43,26 +56,43 @@ private static StatsdReporter provideStatsDHandlers() {
...
             handlers.count("countTest", 1, tags);
             handlers.increment("closeSocket", tags);
-        });
+            log.info("done with handlers");
+            canListen.set(true);
+            return "done";
+        }, executorService);
+        log.info("about to get futures");
+        CompletableFuture.allOf(sender, udpServer.getServerFuture(), udpServerDD.getServerFuture())
+                .get(15000, TimeUnit.MILLISECONDS);
```
manjula-chikkanagappa: I see one more 15000. Can you move this to a constant variable?

src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
manjula-chikkanagappa: Right. Java uses "Pass by value" for method parameters.

src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
```diff
@@ -17,57 +17,66 @@
...
+                    String result = listen(receivePacket);
+                    addValueToStore(result);
+                }
+            } catch (IOException e) {
+                listening.set(false);
             }
-        } catch(Exception ex) {
-            listening = false;
-        }
 //        logger.info(":::::::::, {}", store);
```
manjula-chikkanagappa: why is this commented out?

src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
lkemmerer: I wound up moving all those variables out of the class, so I guess my question is moot. I'll take a look at the concurrency book, though. Thanks!

src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
lkemmerer: Because it was originally and I didn't change it. I can remove it.

src/test/java/com/flightstats/hub/util/IntegrationUdpServer.java
```diff
@@ -1,73 +1,66 @@
...
+                    byte[] data = new byte[70];
+                    DatagramPacket receivePacket = new DatagramPacket(data, data.length);
+                    startupCountDownLatch.countDown();
+                    serverSocket.receive(receivePacket);
+
+                    String result = listen(receivePacket);
+                    addValueToStore(result);
+
+                    if (result.contains("closeSocket")) {
+                        listening = false;
```
chriskessel: Sort of nit picky, but if this is the only way to exit the loop, you can remove listening completely and just ```break``` here.

src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java
lkemmerer: It's a lot because we don't lose that time if things return faster and having something outrageous seemed okish. We were having issues with 5s timeouts at some point in the lifetime of this test. I'll drop it to 5 and make sure things still pass consistently.

src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java
lkemmerer: Technically the two values aren't related, so I'm not sure changing this to a constant makes sense.

src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java
lkemmerer: I'm cool with using a @SneakyThrows here since it's in a test, but if I understand how the annotation works, I think we need to be careful about where we use it. Not using checked exceptions feels like it obfuscates the API of the method.

chriskessel: FWIW, my SneakyThrows usage typically falls into two cases:

* I would have just wrapped it in a RuntimeException anyway.
* I'm only adding it to the method declaration because I have no idea what to do with it. Now, if an upper level can make a smart decision with the checked exception, you want to explicitly have it on the method throws declaration. If nothing in your code base is going to intelligently handle the checked exception, then you might as well SneakyThrows.

lkemmerer: There have definitely been some points in hub code where the latter point applied. It felt weird.

src/test/java/com/flightstats/hub/metrics/StatsdReporterIntegrationTest.java
```diff
@@ -35,13 +36,14 @@
...
+    @SneakyThrows
     @Test
-    public void StatsDHandlersCount_metricShape() throws InterruptedException, ExecutionException, TimeoutException {
+    public void StatsDHandlersCount_metricShape() {
         CompletableFuture.allOf(
                 getMetricsWriterFuture(),
                 udpServer.getServerFuture(startupCountDownLatch, executorService),
                 udpServerDD.getServerFuture(startupCountDownLatch, executorService)
-        ).get(15000, TimeUnit.MILLISECONDS);
+        ).get(5000, TimeUnit.MILLISECONDS);
```
manjula-chikkanagappa: 👍

PR #1156, opened 2019-04-11
Chaos test

manjula-chikkanagappa:


src/systemTest/java/com/flightstats/hub/functional/callback/CallbackCache.java
```diff
@@ -1,6 +1,6 @@
-package com.flightstats.hub.callback;
+package com.flightstats.hub.functional.callback;
 
-import com.flightstats.hub.model.WebhookCallbackRequest;
+import com.flightstats.hub.functional.model.WebhookCallbackRequest;
```
lkemmerer: I think I understand the reasoning for moving models under `functional`, since right now they're only used there, but they could be useful as a package on their own. We could eventually use them + gson in the hub itself so we're not manually constructing JSON (which, iirc, we do!). It could also be used as a dependency in Paul's callback server and other teams have expressed an interest in using something like a model and client library that we provide.

src/systemTest/java/com/flightstats/hub/resilient/HelmReleaseTest.java
```diff
@@ -0,0 +1,35 @@
...
+    @Inject
+    private ReleaseDelete releaseDelete;
+
+    @Inject
+    @Named("helm.chart.path")
+    private String chartPath;
+
+    @Test
+    public void test() {
+        this.releaseInstall.install(this.releaseName, this.chartPath);
```
lkemmerer: If this is going to be a long-lived test, it may make sense to add some sort of assertion that resource creation gets rolling. Does install wait for all the services and whatnot to return good values from their readiness probes? If not, we'll probably need to figure out how to check that various services/sites start up before we start running the rest of our tests.

src/systemTest/java/com/flightstats/hub/functional/callback/CallbackCache.java
manjula-chikkanagappa: I thought about creating a reusable hub client. But am not sure if this is really worth the effort.
All other teams have been using their own hub clients for a while. It would be an additional unrewarding effort for teams to switch to new hub client without any obvious benefits. Given that Hub contract is not going to change anytime in the near future, I would say its not worth it. With Data lake work in progress, there is high possibility that hub may not be used in the future. Again this is something architecture will decide.

src/systemTest/java/com/flightstats/hub/resilient/HelmReleaseTest.java
manjula-chikkanagappa: oh yeah. There is a class ReleaseInstall that ensures this. Again helm package for hub is WIP. Some of the resources fail to instantiate. Rich is working on sequencing the resources instantiation. This will be WIP too.

src/systemTest/java/com/flightstats/hub/functional/callback/CallbackCache.java
lkemmerer: That may be true, but moving the models into the functional tests when it's reasonable that they could be used elsewhere (like the callback server code in one of our other repositories) seems counterintuitive to me. Leaving them outside the functional tests makes it more apparent that we *do* have the beginnings of a shareable model/client library.

manjula-chikkanagappa: ok moved the model package out of functional tests

src/systemTest/java/com/flightstats/hub/resilient/helm/ReleaseInstall.java
```diff
@@ -0,0 +1,54 @@
...
+            final InstallReleaseRequest.Builder requestBuilder = InstallReleaseRequest.newBuilder();
+            requestBuilder.setTimeout(300L);
+            requestBuilder.setName(releaseName);
+            requestBuilder.setWait(true);
+            requestBuilder.setDisableHooks(false);
+            requestBuilder.clearWait();
+
+            final Future<InstallReleaseResponse> releaseFuture = releaseManager.install(requestBuilder, chartBuilder);
+            ReleaseOuterClass.Release release = releaseFuture.get().getRelease();
+            assertTrue(release.hasChart());
```
lkemmerer: I think I didn't realize there were assertions in here when I made the comment about making sure this method works. Did you put them here as a gut-check that things work right every time?

src/systemTest/java/com/flightstats/hub/resilient/testsuite/HelmReleaseTest.java
```diff
@@ -0,0 +1,38 @@
...
+    @Inject
+    private ReleaseInstall releaseInstall;
+
+    @Inject
+    private ReleaseDelete releaseDelete;
+
+    @Test
+    public void test() {
+        this.releaseInstall.install(this.releaseName, this.chartPath);
+        new JUnitCore().run(WebhookLifecycleTest.class);
```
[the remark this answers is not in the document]
lkemmerer: ...anyway, I think we might be able to reconfigure this a bit so that we're not manually running tests inside another test. We don't necessarily need to get fancy with a custom test runner, but it might be worth looking at.

README.md
```diff
@@ -43,7 +43,7 @@ From Local:
     a) ./localDocker.sh (from root directory)
     b) docker run -p 80:80 hub:local
    
-2) set base.url property in integration-hub.properties
+2) set base.url property in functional-hub.properties
```
lkemmerer: The file might already be `system-hub.properties`, but I think that's the name we should use. I think the rest of this readme may need review as well.

manjula-chikkanagappa:
> It might make sense to sit down and chat about this after you read my comments. I'm sure I didn't do a great job conveying my proposal or the reasons behind it.
> 
> As far as my previous comment about keeping functional and resilient tests separate vs keeping everything under resilient... I think merging them isn't a big deal, but if we do we may want to keep them under the main systemTest package instead.

Sure. that would be faster and easier.

src/systemTest/java/com/flightstats/hub/resilient/testsuite/HelmReleaseTest.java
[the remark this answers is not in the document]
manjula-chikkanagappa: @BeforeAll and @AfterAll are setup and tear down methods per class. It is run once before all tests in a class. Currently there is no way of doing setup and teardown for the whole test suite in Junit 5. Yes I am thinking of running multiple testcases. More like this. 1) Run a test. 2) Simulate fault in the cluster. 3) Rerun the test to check if all tests pass.

src/systemTest/java/com/flightstats/hub/resilient/helm/ReleaseInstall.java
manjula-chikkanagappa: Yes. The statement before this (releaseFuture.get()) is only blocked until the job is complete. But does not convey if the chart installation was successful or not.

src/systemTest/java/com/flightstats/hub/resilient/BaseTest.java
```diff
@@ -15,15 +16,15 @@
 @Slf4j
 public class BaseTest {
 
-    protected Injector injector = Guice.createInjector(new GuiceModule());
```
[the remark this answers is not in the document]
manjula-chikkanagappa: I like this approach better. One has to just extend the test from this base class and  all required components for the test is already setup. The base test reads properties file, creates retrofit clients for hub, callback server and sets up Guice DI. The required dependencies in the tests can be injected using annotations. Baseclass is encapsulating all the details from the tests.
Setting up configuration in each test will be redundant.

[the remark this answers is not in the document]
lkemmerer: I think I can meet you halfway and keep this base class. Instead of `BaseTest` call it `GuiceInjection` or something similar...and then move all the protected methods out into another client-related helper class that can be injected into the tests. That way the only behavior in this class would be to handle injection. Limiting behavior plus renaming will reduce the likelihood that folks will put other "helping" behavior in here. Might make sense to remove the @Slf4j annotation, too, so we don't have to override stuff in all the tests.

PR #1158, opened 2019-04-15
Pdxddt 579 s3 monitor

Paul-Hess:


src/main/java/com/flightstats/hub/dao/aws/HubS3Client.java
```diff
@@ -108,6 +108,15 @@ void deleteObject(DeleteObjectRequest request) {
...
     }
 
+    void deleteVersion(DeleteVersionRequest request) {
+        try {
+            s3Client.deleteVersion(request);
+        } catch (SdkClientException e) {
+            countError(e, request, "deleteVersion", Arrays.asList("bucket:" + request.getBucketName(), "key:" + request.getKey()));
+            throw e;
+        }
+    }
```
manjula-chikkanagappa: Why is there a manual delete on s3 object versions? Isn't this managed thru s3 bucket lifecycle policies?

src/test/java/com/flightstats/hub/dao/aws/S3AccessMonitorTest.java
```diff
@@ -0,0 +1,75 @@
...
+        assertFalse(monitor.verifyReadWriteAccess());
+    }
+
+
+    @Test
+    public void testVerifyReadWriteAccess_mockVersionId_true() {
+        HubS3Client s3Client = mock(HubS3Client.class);
+        S3AccessMonitor monitor = new S3AccessMonitor(s3Client, s3BucketName);
+        PutObjectResult putObjectResult = new PutObjectResult();
+        putObjectResult.setVersionId("testVersionId");
```
manjula-chikkanagappa: This is repeated block of code in all tests. this can be moved to @Before junit method

src/main/java/com/flightstats/hub/app/ClusterHubBindings.java
```diff
@@ -4,8 +4,28 @@
...
+import com.flightstats.hub.dao.aws.S3BatchContentDao;
+import com.flightstats.hub.dao.aws.S3LargeContentDao;
+import com.flightstats.hub.dao.aws.S3AccessMonitor;
+import com.flightstats.hub.dao.aws.HubS3Client;
+import com.flightstats.hub.dao.aws.S3BatchManager;
+import com.flightstats.hub.dao.aws.DynamoUtils;
+import com.flightstats.hub.dao.aws.S3Verifier;
+import com.flightstats.hub.dao.aws.S3DocumentationDao;
+import com.flightstats.hub.metrics.PeriodicMetricEmitter;
+import com.flightstats.hub.metrics.PeriodicMetricEmitterLifecycle;
```
manjula-chikkanagappa: 👍

src/main/java/com/flightstats/hub/dao/aws/HubS3Client.java
Paul-Hess: It could be handled through lifecycle policy, but as far as I can tell that would use another of our limited LPs, by immediately deleting the version manually I am trying to avoid any costs of orphaned versions while leaving the LPs to be available for consumer channel TTLs

Paul-Hess:
> I made a comment on deleting version objects. After entire review, I kind of understand that you are testing s3 access. Are you writing some metrics to s3 bucket? Is there a reason for doing this?
.
Reason being any of sudden loss of connectivity to S3, permissions change, auth update, firewall or network issues, amazon outage, etc... 
Not writing the metrics to S3 bucket but to datadog/influxdb for alerts consumption in that context

src/main/java/com/flightstats/hub/app/HubMain.java
```diff
@@ -54,6 +55,7 @@
 public class HubMain {
 
     private static final DateTime startTime = new DateTime();
+    private final String hubType = HubProperties.getProperty("hub.type", "aws");
```
chriskessel: This feels like something that wants to be an enum. Which, for bonus points, an enum.valueOf() will also blow up if you pass it a value it doesn't know.

src/main/java/com/flightstats/hub/dao/aws/S3AccessMonitor.java
```diff
@@ -0,0 +1,92 @@
...
+    private final HubS3Client hubS3Client;
+    private final S3BucketName s3BucketName;
+    private final Content content;
+    private static final String CHANNEL_NAME = "S3_HEALTH_MONITOR";
+
+    @Inject
+    public S3AccessMonitor(HubS3Client hubS3Client, S3BucketName s3BucketName) {
+        this.hubS3Client = hubS3Client;
+        this.s3BucketName = s3BucketName;
+        String data = "1";
```
chriskessel: Magic number "1" could use an self-documenting constant (or rename "data" into something self-documenting).

src/main/java/com/flightstats/hub/dao/aws/S3AccessMonitor.java
```diff
@@ -0,0 +1,92 @@
...
+            });
+        } catch(Exception e) {
+            log.error("error getting object from s3", e);
+            throw e;
+        }
+    }
+
+    private void cleanup(String versionId) {
+        try {
+            DeleteObjectRequest deleteObjectRequest = new DeleteObjectRequest(s3BucketName.getS3BucketName(), key());
```
chriskessel: I'm not familiar with s3 versioning, but do you want to execute the delete both with and without the version or should this be an if/else?

Or maybe a 1-line comment before the if as to why it's in addition to, rather than instead of, the main delete.

src/main/java/com/flightstats/hub/dao/aws/HubS3Client.java
manjula-chikkanagappa: We know that Lifecycle policies offered by S3 bucket is not sufficient to cleanup stale data in S3 bucket. We do need an alternative/additional solution to accomplish this. IMO, it is not good idea to introduce another approach to cleaning up data.

src/main/java/com/flightstats/hub/app/HubMain.java
Paul-Hess: I agree, I will time box a solution that does not use HubProperties, that class is deprecated and slated for change in another story. Changing how it is used here is a pattern that would warrant changes in integral startup processes of the hub

src/main/java/com/flightstats/hub/dao/aws/S3AccessMonitor.java
Paul-Hess: I mistook the docs on the delete version method on first pass, I thought that it only deleted the "delete marker" but it does delete the object as well, updated.

src/main/java/com/flightstats/hub/app/HubMain.java
chriskessel: Longer term, it'd be nice to have something more type safe than HubProperties. All those strings is asking for a tpyo issue, but I know that's scattered throughout the code. However, in this case I was thinking smaller scale, something like this:
```
private final HubType hubType = HubType.valueOf(HubProperties.getProperty("hub.type", "aws"));
```

src/main/java/com/flightstats/hub/dao/aws/HubS3Client.java
Paul-Hess: I feel like this is pretty isolated for an isolated problem, but I see your point

src/main/java/com/flightstats/hub/app/StorageBackend.java
```diff
@@ -0,0 +1,15 @@
+package com.flightstats.hub.app;
+
+public enum StorageBackend {
+    AWS("aws"),
+    NAS("nas"),
+    TEST("test");
```
manjula-chikkanagappa: This is for my understanding. What does these values mean? AWS means S3? NAS means spoke storage?

chriskessel: FWIW, when the value differs from the name only by case, I just use the appropriately cased name. E.g:
```
public enum StorageBackend { aws, nas, test; }
```

Goes a little against the uppercase convention for constants, but I don't think of enum at the same conceptual level as constants. You can even think of the enum values as "public final" instance variables (which aren't upper cases) of the StorageBackend class if that helps.

Mostly, I ask myself "Is this extra string that differs only by case adding value?" If the answer is no, then I ignore the uppercase convention. If the answer is yes, then I tend to uppercase them since the real valuable information is the parameters they're instantiated with.

Paul-Hess: @manjula-chikkanagappa aws == s3, nas == network attached storage (I always forget what that acronym is)

Paul-Hess: glad to hear about the casing convention because the upper-casing was giving me problems in tests

PR #1159, opened 2019-04-17
Fix webhook error state not always cleared on recreate/cursor update

dkleckler:


src/main/java/com/flightstats/hub/cluster/DistributedAsyncLockRunner.java
```diff
@@ -52,11 +52,16 @@ public void setLockPath(String lockPath) {
...
     public void delete(LeadershipLock leadershipLock) {
         try {
-            executorService.awaitTermination(1, TimeUnit.SECONDS);
+            executorService.shutdown();
+            if (!executorService.awaitTermination(1, TimeUnit.SECONDS)) {
+                executorService.shutdownNow();
+                executorService.awaitTermination(3, TimeUnit.SECONDS);
+            }
         } catch (InterruptedException e) {
             log.info("InterruptedException for " + leadershipLock.getLockPath(), e);
```
chriskessel: I know it's not a change in this PR, but you almost always want to call ```Thread.currentThread().interrupt()``` to restore the interrupted flag as part of handling an InterruptedException. Otherwise, whatever spawned the thread (e.g. a thread pool) that invoked this method won't see the interrupt occurred and clean up properly.

The reasons are...really mostly beyond me, but Java Concurrency in Practice goes over it and this Stack Overflow answer does a good job of explaining it.
https://stackoverflow.com/questions/3976344/handling-interruptedexception-in-java

src/systemTest/java/com/flightstats/hub/callback/CallbackResource.java
```diff
@@ -13,23 +14,39 @@
...
 @Path("/callback")
 public class CallbackResource {
 
     private CallbackCache callbackCache;
+    private List<Predicate<WebhookCallbackRequest>> responseOverrides;
 
     @Inject
     public CallbackResource(CallbackCache callbackCache) {
         this.callbackCache = callbackCache;
+        this.responseOverrides = Collections.synchronizedList(new ArrayList<>());
```
chriskessel: If thread safe collections are important, I highly recommend using java.util.concurrent classes. In this case, CopyOnWriteArrayList.

src/systemTest/java/com/flightstats/hub/callback/CallbackResource.java
```diff
@@ -13,23 +14,39 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Predicate;
 
 @Slf4j
 @Path("/callback")
 public class CallbackResource {
 
     private CallbackCache callbackCache;
```
chriskessel: These variables can be final?

src/main/java/com/flightstats/hub/cluster/DistributedAsyncLockRunner.java
dkleckler: Yeah, agreed. I avoided adding it just to keep the changeset minimal since I was messing with other threading aspects. I'll add it back in and see if it seems copacetic.  Interestingly, we only reset this in 1/13 occurrences in hub currently.

chriskessel: It's one of those things that's bad, but you can get away with forgetting it...mostly. It typically comes into play on shutdown of a Executor (thread pool) which usually happens when you're terminating a process. Forgetting to restore the interrupt can mean cleanup doesn't happen as gracefully as it should, which may or may not be a big deal depending on the application.

src/systemTest/java/com/flightstats/hub/callback/CallbackResource.java
```diff
@@ -13,23 +14,39 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Predicate;
 
 @Slf4j
 @Path("/callback")
 public class CallbackResource {
 
-    private CallbackCache callbackCache;
+    private final CallbackCache callbackCache;
+    private final List<Predicate<WebhookCallbackRequest>> responseOverrides;
```
Paul-Hess: Wanted to note that in #1156 this file is removed, I am (just assuming) that the intended use here will be moved to the external callback server in our k8s tests. You may have to coordinate with @manjula-chikkanagappa  to confirm, this error cache functionality looks good generally and is something that has not been added to that repo yet, which you can PR or let me know if you are busy with other stuff and I will add it

dkleckler: Yeah agreed.  I think this is something @manjula-chikkanagappa is starting to add on her side.  Pinged her to get her thoughts on it.

src/systemTest/java/com/flightstats/hub/testcase/WebhookErrorTest.java
```diff
@@ -0,0 +1,136 @@
...
+import org.slf4j.Logger;
+
+import java.util.Optional;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+@Slf4j
+public class WebhookErrorTest extends WebhookTest {
```
lkemmerer: I have a pretty staunchly held belief that inheritance is almost always the wrong way to go when structuring code. This is the subclass of a subclass, so I feel like the inheritance is getting too deep (and I have a suggestion in another PR for getting rid of BaseTest as a base class). I'd split WebhookTest into a few classes (webhooks, callbacks, channel, maybe one for shared behavior) and instantiate them here.

src/systemTest/java/com/flightstats/hub/testcase/WebhookTest.java
```diff
@@ -0,0 +1,221 @@
...
+        log.info("Call back errors for webhook {} {} ", webhookName, body);
+    }
+
+    @SneakyThrows
+    protected Response<WebhookErrors> getCallbackErrorsFromHub() {
+        Call<WebhookErrors> call = webhookResourceClient.getError(webhookName);
+        Response<WebhookErrors> response = call.execute();
+
+        WebhookErrors body = response.body();
+        assertEquals(OK.getStatusCode(), response.code());
```
lkemmerer: If these assertions are gut-checks, I think it's fine to keep these here. If not, I think moving the assertions into a different method that calls out that they assert something may make sense.

src/systemTest/java/com/flightstats/hub/testcase/WebhookErrorTest.java
```diff
@@ -0,0 +1,136 @@
...
+        // verify that no errors exist on the hub
+        log.info("Verifying that no errors exist on the hub for webhook {}", webhookName);
+        assertFalse(super.hasCallbackErrorInHub(webhookName, firstUrl));
+    }
+
+    @Test
+    @SneakyThrows
+    public void testWebhookCursorUpdateLoopDoesntCorruptState() {
+
+        for (int i = 1; i <= 5; i++) {
```
lkemmerer: Was 5 selected for a reason? Would 2 or 3 be enough?

dkleckler: 3 would probably be enough.

src/systemTest/java/com/flightstats/hub/testcase/WebhookErrorTest.java
dkleckler: Yeah, I think the changes in Manjula's k8s branch will require significant re-writes to this, so I didn't spend much time on the design here.

lkemmerer: Fair enough. I'll watch for the refactoring in another PR.

manjula-chikkanagappa: I agree with Laurie here. Multi level inheritance can make things complex and is error prone. I would consider adding helper classes here.

src/systemTest/java/com/flightstats/hub/testcase/WebhookTest.java
manjula-chikkanagappa: ok. This is coz the response body is logged in the next statement. If for some reason the service call fails, it will result in NPE. Assertion is way of logging the response code from the service call in case of error.

src/systemTest/java/com/flightstats/hub/testcase/WebhookErrorTest.java
dkleckler: Yep, I agree.  I can refactor this if you'd like, it just seemed not worth investing much in since it's going to change so much once your branch comes in.

manjula-chikkanagappa: Don't worry about my branch. Go head. I will deal with it.

PR #1160, opened 2019-04-17
Refactor statsd test

Paul-Hess:


Paul-Hess:
I don't know exactly why, but it was losing the "closeSocket" packet. I put a lot of logging in between the first commit and the last that confirmed this, no longer visible because I squashed all those commits.

PR #1161, opened 2019-04-18
Use timing method

Paul-Hess:
https://jira.flightglobal.io/browse/PDXDDT-134

edited with updated datadog link:

https://app.datadoghq.com/dashboard/wsq-d9n-26w/hubs-by-cluster?tile_size=l&page=0&is_auto=false&from_ts=1555621395000&to_ts=1555622295000&live=true&tpl_var_cluster=dls.dev&fullscreen_widget=3543467067562178

### DIFF
--- a/src/main/java/com/flightstats/hub/dao/aws/S3WriteQueue.java
+++ b/src/main/java/com/flightstats/hub/dao/aws/S3WriteQueue.java
@@ -69,7 +69,7 @@ public class S3WriteQueue {
             ChannelContentKey key = keys.poll(5, TimeUnit.SECONDS);
             if (key != null) {
                 statsdReporter.gauge("s3.writeQueue.used", keys.size());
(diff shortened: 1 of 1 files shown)


lkemmerer:
Dunno if it'd make sense to keep both metrics around until we decide which one makes more sense, but if you feel like this is definitely the right metric, I'm cool with that, too

lkemmerer:
...nevermind, for some reason I forgot you can test this in dev, despite the link. :P

Paul-Hess:
the link to the Datadog graph is updated and I think displays the intended use of both metrics together. To get a similar graphing I think the hub would have to do more work that I am fine with handing off to the <influxdb|datadog> query

lkemmerer:
Makes sense. I am a little weirded out that the time spent in the queue is 0, but it might just be that it's a relatively low-volume dev cluster.

Paul-Hess:
I believe that the oddly low metrics are because it is deployed on dls dev

PR #1162, opened 2019-04-19
Refactor systemTest to move inherited capabilities out to helper classes.

dkleckler:


lkemmerer:
Since this is merged already, I guess these comments are for a PR after Manjula merges her work. Suggestions:
- Rename BaseTest to be more descriptive of its single purpose (allowing guice injection).
- Rename Helpers classes to something slightly more descriptive (Actions? I can't think of a good name, but I think helper is too vague).
- Split HubHelper out into webhook and channel (and maybe split out actions from assertions if there's clean separation). If the file is 140 lines after a couple of tests, that's gonna be a monster class.

PR #1163, opened 2019-04-20
remove drop some writes and some of its side effects

Paul-Hess:


lkemmerer:
Oh! Does it need to be removed from any properties/config files?

manjula-chikkanagappa:
Yes. I will take care of this in PDXDDT-635 Hub config clean up

Paul-Hess:
I believe it is gone from all config files already

PR #1164, opened 2019-04-22
ensure get object stream is closed

Paul-Hess:


dkleckler:
It would be nice to have an explicit test that S3Object.close() gets called regardless of success or failure.

src/main/java/com/flightstats/hub/dao/aws/S3AccessMonitor.java
```diff
@@ -89,6 +89,7 @@ private String key() {
                     return s3Object.getObjectMetadata().getVersionId();
                 } catch (IOException e) {
                     log.info("error closing connection to s3", e);
+                    return "";
```
manjula-chikkanagappa: It would be nice to use constant here StringUtils.EMPTY or declare it in the class. This prevents the accidental modification of the "".

PR #1165, opened 2019-04-24
S3 write queue tests

Paul-Hess:
I wanted to write some basic S3WriteQueue unit tests, particularly to document/test behavior that dropSomeWrites was previously in the code base to test. I got deeper in pulling at threads than I would have hoped so now I am putting it up for discussion before it goes deeper. Ultimately, I know that @manjula-chikkanagappa is working on config stuff that may supersede some of this refactoring and @dkleckler is working on integration refactoring around this as well. I'll be happy to wait for those PRs to go through first and merge whatever is left after some discussion and rebasing to those PRs.

### DIFF
--- a/src/main/java/com/flightstats/hub/app/ShutdownManager.java
+++ b/src/main/java/com/flightstats/hub/app/ShutdownManager.java
@@ -1,5 +1,6 @@
 package com.flightstats.hub.app;
 
+import com.flightstats.hub.dao.aws.S3WriteQueue;
 import com.flightstats.hub.health.HubHealthCheck;
 import com.flightstats.hub.metrics.StatsdReporter;
 import com.flightstats.hub.util.Sleeper;
@@ -62,6 +63,8 @@ public class ShutdownManager {
         }
 
         HubServices.stopAll();
+        S3WriteQueue s3WriteQueue = HubProvider.getInstance(S3WriteQueue.class);
+        s3WriteQueue.close();
         log.warn("completed shutdown tasks, exiting JVM");
         Executors.newSingleThreadExecutor().submit(() -> System.exit(0));
         return true;
--- a/src/main/java/com/flightstats/hub/app/ShutdownManager.java
+++ b/src/main/java/com/flightstats/hub/app/ShutdownManager.java
@@ -63,8 +63,6 @@ public class ShutdownManager {
         }
 
         HubServices.stopAll();
-        S3WriteQueue s3WriteQueue = HubProvider.getInstance(S3WriteQueue.class);
-        s3WriteQueue.close();
         log.warn("completed shutdown tasks, exiting JVM");
         Executors.newSingleThreadExecutor().submit(() -> System.exit(0));
         return true;
--- a/src/main/java/com/flightstats/hub/app/ShutdownManager.java
+++ b/src/main/java/com/flightstats/hub/app/ShutdownManager.java
@@ -1,6 +1,5 @@
 package com.flightstats.hub.app;
 
-import com.flightstats.hub.dao.aws.S3WriteQueue;
 import com.flightstats.hub.health.HubHealthCheck;
 import com.flightstats.hub.metrics.StatsdReporter;
 import com.flightstats.hub.util.Sleeper;
@@ -10,8 +9,6 @@ import com.google.inject.Singleton;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.zookeeper.KeeperException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.util.concurrent.Executors;
 
--- a/src/main/java/com/flightstats/hub/dao/aws/S3WriteQueue.java
+++ b/src/main/java/com/flightstats/hub/dao/aws/S3WriteQueue.java
@@ -7,21 +7,17 @@ import com.flightstats.hub.metrics.ActiveTraces;
 import com.flightstats.hub.metrics.StatsdReporter;
 import com.flightstats.hub.model.ChannelContentKey;
 import com.flightstats.hub.model.Content;
-import com.flightstats.hub.util.Sleeper;
 import com.github.rholder.retry.Retryer;
 import com.github.rholder.retry.RetryerBuilder;
 import com.github.rholder.retry.StopStrategies;
 import com.github.rholder.retry.WaitStrategies;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.inject.Singleton;
 import com.google.inject.name.Named;
 import lombok.extern.slf4j.Slf4j;
 
 import javax.inject.Inject;
 import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
@@ -33,9 +29,7 @@ public class S3WriteQueue {
 
     @VisibleForTesting
     BlockingQueue<ChannelContentKey> keys;
-    private ExecutorService executorService;
 
-    private final WriteQueueConfig writeQueueConfig;
     private final ContentDao spokeWriteContentDao;
     private final ContentDao s3SingleContentDao;
     private final StatsdReporter statsdReporter;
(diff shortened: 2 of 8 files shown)


src/main/java/com/flightstats/hub/dao/aws/S3WriteQueue.java
```diff
@@ -47,24 +44,30 @@
     @Inject
     private StatsdReporter statsdReporter;
 
+    S3WriteQueue(ContentDao spokeWriteContentDao, ContentDao s3SingleContentDao, StatsdReporter statsdReporter) {
```
lkemmerer: I think you might be able to annotate this with `@Inject` and add some `@Named` annotations to the content daos. Then you could make the dependencies `private final`...

src/main/java/com/flightstats/hub/dao/aws/S3WriteQueue.java
```diff
@@ -47,24 +44,30 @@
...
-                    logger.warn("exited thread", e);
+                    log.warn("exited thread", e);
                     return null;
                 }
             });
         }
     }
 
-    private void write() throws InterruptedException {
+    void write() throws InterruptedException {
```
lkemmerer: If this is no longer private because of testing, I tend to annotate this type of thing with `@VisibleForTesting`

src/test/java/com/flightstats/hub/dao/aws/S3WriteQueueTest.java
```diff
@@ -0,0 +1,110 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@Slf4j
+public class S3WriteQueueTest {
+
+    private SpokeWriteContentDao spokeWriteContentDao = mock(SpokeWriteContentDao.class);
```
lkemmerer: Damon showed me a `@Mock` annotation to use here instead of explicitly calling `mock()` which I thought was pretty cool. :shrug:

dkleckler: It simplifies things a bit, I think.  ClusterContentServiceTest has this, if you're interested.

src/main/java/com/flightstats/hub/dao/aws/S3WriteQueue.java
```diff
@@ -14,57 +13,62 @@
...
+        this.spokeWriteContentDao = spokeWriteContentDao;
+        this.s3SingleContentDao = s3SingleContentDao;
+        this.statsdReporter = statsdReporter;
+        this.writeQueueConfig = writeQueueConfig;
+        keys = new LinkedBlockingQueue<>(writeQueueConfig.getQueueSize());
+        executorService = Executors.newFixedThreadPool(writeQueueConfig.getThreads(),
+                new ThreadFactoryBuilder().setNameFormat("S3WriteQueue-%d").build());
+    }
+
+    public void start() throws InterruptedException {
```
Paul-Hess: I think this should be refactored into an AbstractService https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/util/concurrent/AbstractService.html 
We are using that pattern throughout the rest of the hub for similar long running "services"
Would like feedback before I went further on that.

dkleckler:
Most of my changes are in *Bindings and ClusterContentService so far.  This stuff shouldn't conflict with what I'm working on unless you try to decouple S3WQ from spoke.  I'll merge it over to my branch once it hits master.

src/main/java/com/flightstats/hub/app/ShutdownManager.java
```diff
@@ -62,6 +63,8 @@ public boolean shutdown(boolean useLock) throws Exception {
         }
 
         HubServices.stopAll();
+        S3WriteQueue s3WriteQueue = HubProvider.getInstance(S3WriteQueue.class);
+        s3WriteQueue.close();
```
Paul-Hess: This method was not being called before. Not sure it is needed. If we make S3WriteQueue an AbstractService as suggested below the ServiceManager class would handle this

src/main/java/com/flightstats/hub/dao/aws/S3WriteQueue.java
Paul-Hess: that was done in the latest commit

src/main/java/com/flightstats/hub/dao/aws/S3WriteQueue.java
Paul-Hess: Yes, I can annotate it

src/test/java/com/flightstats/hub/dao/aws/S3WriteQueueTest.java
Paul-Hess: 👍

src/main/java/com/flightstats/hub/dao/aws/writeQueue/WriteQueueConfigProvider.java
```diff
@@ -0,0 +1,13 @@
+package com.flightstats.hub.dao.aws.writeQueue;
+
+import com.flightstats.hub.app.HubProperties;
+import com.google.inject.Provider;
+
+public class WriteQueueConfigProvider implements Provider<WriteQueueConfig> {
```
lkemmerer: I personally like this pattern (I may have done something similar somewhere...) and even if there only ends up being two attributes here, I think it's an acceptable size. 👍

src/main/java/com/flightstats/hub/dao/aws/S3WriteQueue.java
lkemmerer: I like that idea, especially if we break with the pattern of doing it as a nested class, if possible.

src/main/java/com/flightstats/hub/app/ShutdownManager.java
lkemmerer: I wonder why it wasn't being called by accident or by design (i.e. it was called and they stopped but never removed the code). Either way, I like the idea of dealing with all this as a service.

src/main/java/com/flightstats/hub/dao/aws/S3WriteQueueLifecycle.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+@Slf4j
+public class S3WriteQueueLifecycle extends AbstractService {
+    private final S3WriteQueue s3WriteQueue;
+    private final WriteQueueConfig writeQueueConfig;
+    private ExecutorService executorService;
+    private AtomicBoolean started = new AtomicBoolean(false);
```
lkemmerer: I thiiiiink this can be final, no?

Paul-Hess: oops yeah

src/test/java/com/flightstats/hub/dao/aws/S3WriteQueueTest.java
```diff
@@ -0,0 +1,118 @@
...
+import static org.mockito.MockitoAnnotations.initMocks;
+
+@Slf4j
+public class S3WriteQueueTest {
+
+    private @Mock SpokeWriteContentDao spokeWriteContentDao;
+    private @Mock S3SingleContentDao s3SingleContentDao;
+    private @Mock StatsdReporter statsdReporter;
+    private @Mock ChannelContentKey key;
+    private @Mock ContentKey contentKey;
```
manjula-chikkanagappa: Hmm.  @Mock should be before private keyword. Here is the example https://static.javadoc.io/org.mockito/mockito-core/2.27.0/org/mockito/Mock.html
The order for class modifiers as per JLS is here https://docs.oracle.com/javase/specs/jls/se11/html/jls-8.html#jls-8.1.1

src/test/java/com/flightstats/hub/dao/aws/S3WriteQueueTest.java
```diff
@@ -0,0 +1,118 @@
...
+    private WriteQueueConfig writeQueueConfig = WriteQueueConfig.builder()
+            .queueSize(20)
+            .threads(2)
+            .build();
+    private final static String CHANNEL_NAME = "testy_test";
+    private final static long AGE_MILLIS = 666;
```
manjula-chikkanagappa: same here. The order of modifiers should be public static final 
https://docs.oracle.com/javase/specs/jls/se11/html/jls-8.html#jls-8.1.1

src/test/java/com/flightstats/hub/dao/aws/S3WriteQueueTest.java
```diff
@@ -0,0 +1,118 @@
...
+    public void setup() {
+        initMocks(this);
+        s3WriteQueue = new S3WriteQueue(
+                spokeWriteContentDao,
+                s3SingleContentDao,
+                statsdReporter,
+                writeQueueConfig);
+    }
+
+    @Test
```
manjula-chikkanagappa: can we start using junit 5 for new tests/ updates?

Paul-Hess: junit 5 is not currently installed in the repo, build.gradle: `testCompile 'junit:junit:4.+'` 
I still have a branch to introduce it, but other things have been prioritized.

PR #1167, opened 2019-04-30
Config module

manjula-chikkanagappa:


lkemmerer:
@manjula-chikkanagappa What are your thoughts on leveraging Guice and `Names.bindProperties` to inject configuration properties into our DI framework? It'd get rid of the awkward `PropertyLoader.getInstance()` method calls... 
I was also hoping to use Guice Providers to build up Lombok'ed config POJOs so that logic and defaults could be centralized and tested more explicitly. `MetricsConfig` and `MetricsConfigProvider` kind of go that route without having done the property guicification.
 ...I'm not sure how much static hub weirdness you'd have to overcome with that approach. ...guice does provide a crutch to help deal with that, at least...

manjula-chikkanagappa:
> @manjula-chikkanagappa What are your thoughts on leveraging Guice and `Names.bindProperties` to inject configuration properties into our DI framework? It'd get rid of the awkward `PropertyLoader.getInstance()` method calls...
> I was also hoping to use Guice Providers to build up Lombok'ed config POJOs so that logic and defaults could be centralized and tested more explicitly. `MetricsConfig` and `MetricsConfigProvider` kind of go that route without having done the property guicification.
> ...I'm not sure how much static hub weirdness you'd have to overcome with that approach. ...guice does provide a crutch to help deal with that, at least...

Yep I know. As much as I would like to use @Inject, some of the classes are instantiated manually. These thing will have to to be fixed eventually. One step at a time. Since these changes span all over the hub app, I would like limit the number of changes and also its impact on unit and integration tests. Also the approach I have taken to modularize the config properties is much simpler and also reduces the number of classes.

src/main/java/com/flightstats/hub/webhook/LocalWebhookManager.java
```diff
@@ -22,38 +21,44 @@
...
-    private Map<String, WebhookLeader> localLeaders = new ConcurrentHashMap<>();
+
+    private static int shutdownThreadCount;
 
     @Inject
-    public LocalWebhookManager() {
+    public LocalWebhookManager(@Named("Webhook") Dao<Webhook> webhookDao,
+                                   Provider<WebhookLeader> v2Provider,
+                               WebhookProperty webhookProperty) {
+        shutdownThreadCount = webhookProperty.getShutdownThreadCount();
```
lkemmerer: ...does this mean that `runAndWait` can be called with a shutdown thread count of 0 if an instance is never instantiated? I'm not sure what the implications of that are, but it seems like it could cause problems.

src/test/java/com/flightstats/hub/channel/ChannelValidatorTest.java
```diff
@@ -31,7 +31,7 @@ public void setUp() throws Exception {
         channelService = mock(ChannelService.class);
         validator = new ChannelValidator(channelService);
         when(channelService.channelExists(any(String.class))).thenReturn(false);
-        HubProperties.setProperty("hub.protect.channels", "false");
+        PropertyLoader.getInstance().setProperty("hub.protect.channels", "false");
```
lkemmerer: It'll be nice when we can move away from this mutable stuff and maybe use mocking, since this will carry over between tests, I think.

src/test/java/com/flightstats/hub/cluster/LastContentPathTest.java
```diff
@@ -26,8 +28,8 @@ public static void setUpClass() throws Exception {
     }
 
     @Before
-    public void setUp() throws Exception {
-        lastContentPath = new LastContentPath(curator);
+    public void setUp() {
+        lastContentPath = new LastContentPath(curator, new AppProperty(PropertyLoader.getInstance()));
```
lkemmerer: Would it be possible to mock AppProperty here instead? ...totally understandable (though sad) if this is actually an integration test, but if not...

src/test/java/com/flightstats/hub/dao/aws/S3AccessMonitorTest.java
```diff
@@ -5,23 +5,28 @@
...
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
-import static org.junit.Assert.assertFalse;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 public class S3AccessMonitorTest {
-    private final S3BucketName s3BucketName = new S3BucketName("test", "bucket");
+    private final S3BucketName s3BucketName = new S3BucketName(
```
lkemmerer: same question about mockability here. Looks like we're mocking the hub s3 client, so I have hope that it's a reasonable thing here.

src/test/java/com/flightstats/hub/dao/aws/s3Verifier/VerifierRangeLookupTest.java
```diff
@@ -10,21 +11,22 @@
...
 
 import static com.flightstats.hub.dao.ChannelService.REPLICATED_LAST_UPDATED;
 import static com.flightstats.hub.dao.aws.S3Verifier.LAST_SINGLE_VERIFIED;
 import static org.junit.Assert.assertEquals;
 
+@Slf4j
 public class VerifierRangeLookupTest {
-    private final static Logger logger = LoggerFactory.getLogger(VerifierRangeLookupTest.class);
+
+    private static final SpokeProperty spokeProperty = new SpokeProperty(PropertyLoader.getInstance());
```
lkemmerer: looks like it might be possible to declare this in the setupClass method, but I might be missing something.

src/test/java/com/flightstats/hub/metrics/TopSetTest.java
```diff
@@ -41,7 +43,7 @@ public void testSingle() {
 
     private void addTraces(int count, long start, TopSortedSet topSortedSet) {
         for (int i = 0; i < count; i++) {
-            Traces traces = new Traces(i);
+            Traces traces = new Traces(new AppProperty(PropertyLoader.getInstance()), i);
```
lkemmerer: this might be another candidate for mocking...

src/test/java/com/flightstats/hub/spoke/FileSpokeStoreTest.java
```diff
@@ -20,32 +20,39 @@
...
     private FileSpokeStore spokeStore;
-    private static final int ttlMinutes = HubProperties.getSpokeTtlMinutes(SpokeStore.WRITE);
+    private String tempDir;
 
     @Before
-    public void setUp() throws Exception {
+    public void setUp() {
         tempDir = Files.createTempDir().getPath();
-        HubProperties.setProperty("spoke.write.path", tempDir);
+        PropertyLoader.getInstance().setProperty("spoke.write.path", tempDir);
```
lkemmerer: Might be another good candidate for mocking so we can avoid mutating test-suite-wide state...

docs/pages/mydoc/hub_properties.md
```diff
@@ -0,0 +1,135 @@
...
+s3.writeQueueThreads=20
+s3.maxConnections=50
+s3.connectionTimeout=10 * 1000
+s3.socketTimeout=30 * 1000
+s3Verifier.run=true
+s3.maxQueryItems=1000
+s3Verifier.baseTimeoutMinutes=2
+s3Verifier.offsetMinutes=15
+s3Verifier.channelThreads=3
+channel.enforceTTL=false
```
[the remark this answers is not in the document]
manjula-chikkanagappa: Yes it is used in single hub mode.  We need to work on modularizing the code that is used only in single hub mode.

docs/pages/mydoc/hub_properties.md
```diff
@@ -0,0 +1,135 @@
...
+tags: []
+summary: 
+sidebar: mydoc_sidebar
+permalink: hub_properties.html
+folder: hub
+---
+
+# Here is the complete list of hub properties that is configurable.
+
+#App Properties
```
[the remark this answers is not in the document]
manjula-chikkanagappa: I know. this is a WIP. This is a consolidated list of properties in hub now. Only a fraction of them is in hub.properties. I added a task to the story

src/main/java/com/flightstats/hub/app/HubHost.java
```diff
@@ -1,20 +1,22 @@
...
-import org.slf4j.LoggerFactory;
+import com.flightstats.hub.config.AppProperty;
+import com.flightstats.hub.config.PropertyLoader;
+import lombok.extern.slf4j.Slf4j;
 
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 
+@Slf4j
 public class HubHost {
```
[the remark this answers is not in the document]
manjula-chikkanagappa: This would be cumbersome since it is used all over Hub. I would do this as a separate task. This  will also include getting rid of deprecated HubProvider class

manjula-chikkanagappa:
> Wow, that was a lot of code to change; nice job sticking with it! After looking at the code more, I agree that your way vs the Provider + POJO is generally more elegant.
> 
> One thing I noticed and started commenting on before I realized there'd be a lot of the same comments is that there are a large number of places where the variables for injected dependencies should be `final`. If you want a hand finding and changing them, let me know. I know this required a lot of rote and boring changes and I'm happy to share the work.
> 
> Once this is in, I'd like to figure out a way to get rid of the PropertyLoader.getInstance thing. I know it means we'll need to refactor away a lot of static methods, but I think it'll be worth it. I also _really_ want to get rid of `setProperty`. Having app-wide mutable state like that has only ever brought trouble in my experience.



> Wow, that was a lot of code to change; nice job sticking with it! After looking at the code more, I agree that your way vs the Provider + POJO is generally more elegant.
> 
> One thing I noticed and started commenting on before I realized there'd be a lot of the same comments is that there are a large number of places where the variables for injected dependencies should be `final`. If you want a hand finding and changing them, let me know. I know this required a lot of rote and boring changes and I'm happy to share the work.
> 
> Once this is in, I'd like to figure out a way to get rid of the PropertyLoader.getInstance thing. I know it means we'll need to refactor away a lot of static methods, but I think it'll be worth it. I also _really_ want to get rid of `setProperty`. Having app-wide mutable state like that has only ever brought trouble in my experience.

	I suggest we make these config related changes in phases. The reduces the scope of testing,  simplify the debugging and rollbacks if required. PropertyLoader.getInstance and setproperty was introduced to support existing code. Once we get rid of static classes and remove deprecated HubProvider classes, hopefully we will no longer need these hacks.
I have created Jira tasks to accomplish these tasks in phases. Please feel free to add to it.
I have addressed some of your comments like renaming classes to plural (properties), making variables final, segregating the properties.
Anything else I would take up in a separate task.

lkemmerer:
For sure. I didn't expect to change any of the PropertyLoader behavior in this PR, but wanted to make sure it gets tracked and done eventually. 👍 
I'll take a look at the tasks. Thanks!

src/main/java/com/flightstats/hub/config/WebhookProperties.java
```diff
@@ -0,0 +1,38 @@
...
+public class WebhookProperties {
+
+    private final PropertiesLoader propertiesLoader;
+
+    @Inject
+    public WebhookProperties(PropertiesLoader propertiesLoader) {
+        this.propertiesLoader = propertiesLoader;
+    }
+
+    public int getCallbackTimeoutMin() {
```
Paul-Hess: I think this name is somewhat ambiguous, when I see "Min" I think minutes not minimum, I would prefer something like "getCallbackTimeoutMinimumInSec" and think it would be consistent to add the time unit to the max get method that follows

src/main/java/com/flightstats/hub/app/HubMain.java
```diff
@@ -57,13 +61,15 @@
 public class HubMain {
 
     private static final DateTime startTime = new DateTime();
-    private final StorageBackend storageBackend = StorageBackend.valueOf(HubProperties.getProperty("hub.type", "aws"));
+    private static PropertiesLoader propertiesLoader = PropertiesLoader.getInstance();
+    private final StorageBackend storageBackend = StorageBackend.valueOf(propertiesLoader.getProperty("hub.type", "aws"));
```
Paul-Hess: I would like to see the usage of "hub.type" refactored to not reference aws but be something more like single, clustered, encrypted, readOnly, etc... and have the aws designation be called app.storage.backend or something. But that is definitely a separate PR

src/main/java/com/flightstats/hub/webhook/LocalWebhookManager.java
```diff
@@ -22,38 +21,44 @@
...
-    private Map<String, WebhookLeader> localLeaders = new ConcurrentHashMap<>();
+
+    private static int shutdownThreadCount;
 
     @Inject
-    public LocalWebhookManager() {
+    public LocalWebhookManager(@Named("Webhook") Dao<Webhook> webhookDao,
+                                   Provider<WebhookLeader> v2Provider,
+                               WebhookProperties webhookProperties) {
+        shutdownThreadCount = webhookProperties.getShutdownThreadCount();
```
lkemmerer: I don't think it's safe to set this static variable in the initializer. It's possible to call `runAndWait` prior to an instance being created, which would set the thread count to 0...unless it's bound as an eager singleton when the app's set up, which I'd rather not depend on.

src/main/java/com/flightstats/hub/config/WebhookProperties.java
lkemmerer: I might add a task somewhere (that'll likely be deprioritized) to use Duration types for properties that have time units coded into their name. sot this'd end up being `Duration getCallbackTmeoutMinimum`. I've found that using the right object type for the value generally DRYs up a lot of the places that end up doing time-based calculations with the method.

lkemmerer:
I've gone through and resolved the comments that I've made that have either been addressed or have tasks added to fix 'em. I'll add some more subtasks and resolve more comments with links to the tasks shortly.

src/main/java/com/flightstats/hub/model/ChannelConfig.java
```diff
@@ -99,7 +101,7 @@ private ChannelConfig(String name,
         addTagIf(!isBlank(replicationSource), REPLICATED);
         addTagIf(isHistorical(), HISTORICAL);
 
-        if (HubProperties.isProtected()) {
+        if (appProperty.isProtected()) {
```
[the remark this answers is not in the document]
lkemmerer: I still don't like this very much, but feel like cleaning it up so this is a dumb object that doesn't require knowledge of app properties would require more refactoring than it's worth. Resolving this without an associated follow-up task.

src/main/java/com/flightstats/hub/dao/aws/S3LargeContentDao.java
```diff
@@ -45,31 +45,33 @@
 
 @SuppressWarnings("Duplicates")
 @Singleton
+@Slf4j
 public class S3LargeContentDao implements ContentDao {
```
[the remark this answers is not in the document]
lkemmerer: PDXDDT-799 tracks this.

src/main/java/com/flightstats/hub/dao/aws/S3LargeContentDao.java
```diff
@@ -273,8 +275,12 @@ public void delete(String channel) {
         private StatsdReporter statsdReporter;
         private HubS3Client s3Client;
         private S3BucketName s3BucketName;
+        private AppProperty appProperty;
+        private S3Property s3Property;
 
-        S3LargeContentDaoBuilder() {
+        S3LargeContentDaoBuilder(AppProperty appProperty, S3Property s3Property) {
```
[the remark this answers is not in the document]
lkemmerer: PDXDDT-799 tracks this.

src/main/java/com/flightstats/hub/dao/aws/s3verifier/VerifierConfigProvider.java
```diff
@@ -1,29 +1,33 @@
...
 
-import com.flightstats.hub.app.HubProperties;
+import com.flightstats.hub.config.AppProperty;
+import com.flightstats.hub.config.S3Property;
 import com.google.inject.Provider;
 
+import javax.inject.Inject;
 import java.util.concurrent.TimeUnit;
 
 public class VerifierConfigProvider implements Provider<VerifierConfig> {
```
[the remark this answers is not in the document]
lkemmerer: PDXDDT-800 tracks this.

src/main/java/com/flightstats/hub/metrics/MetricsConfigProvider.java
```diff
@@ -1,45 +1,54 @@
...
-import com.google.inject.Inject;
+import com.flightstats.hub.config.AppProperty;
+import com.flightstats.hub.config.DatadogMetricsProperty;
+import com.flightstats.hub.config.TickMetricsProperty;
 import com.google.inject.Provider;
 
+import javax.inject.Inject;
+
 public class MetricsConfigProvider implements Provider<MetricsConfig> {
+
```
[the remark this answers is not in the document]
lkemmerer: PDXDDT-800 tracks moving this over to the new pattern. I guess when that's tackled, the usefulness of this class as an umbrella config can be reexamined.

src/main/java/com/flightstats/hub/dao/file/FileUtil.java
```diff
@@ -16,12 +16,13 @@
...
 import java.util.function.Function;
 
+@Slf4j
 class FileUtil {
 
-    private final static Logger logger = LoggerFactory.getLogger(FileUtil.class);
+    private static final SpokeProperty spokeProperty = new SpokeProperty(PropertyLoader.getInstance());
 
     static String getStoragePath() {
-        return StringUtils.appendIfMissing(HubProperties.getProperty("storage.path", "/file"), "/");
```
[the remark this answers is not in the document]
lkemmerer: PDXDDT-801 tracks this.

manjula-chikkanagappa:
> Some of these are repeat comments and some are new but based on code that hasn't changed since your last round of feedback changes. Sorry for both...it's a lot of code and I think I lost track of what I've already commented on. :-/ I think most of it is tiny renames. The only non-trivial changes that I can think of are:
> 
> 1. Making sure all the new Properties classes are bound as eager singletons
> 2. Checking out the implications of loading the properties instance in `HubMain` as a static without the app's arguments.
> 3. Sorting out how to avoid instantiating a static variable in a constructor in `LocalWebhookManager`.

LocalWebhookManager should be handled as part of task that will remove the deprecated HubProvider class.

lkemmerer:
> LocalWebhookManager should be handled as part of task that will remove the deprecated HubProvider class.

I think it should be handled here as there's a non-zero chance it's introducing a new bug or at least a changing previous application behavior.

manjula-chikkanagappa:
> > LocalWebhookManager should be handled as part of task that will remove the deprecated HubProvider class.
> 
> I think it should be handled here as there's a non-zero chance it's introducing a new bug or at least a changing previous application behavior.

Not anymore :) . There is no constructor in LocalWebhookManager!

lkemmerer:
Oh, cool. I didn't see that you'd pushed a change. Thanks!

PR #1170, opened 2019-05-03
Support setting HUB_RMI_HOST_IP as an environment variable when launc…

dkleckler:
…hing docker to get it to expose its JMX controls at that IP:8888.  You must bind 8888 to the same port on the host box.  Authentication is NOT enabled, so probably avoid this in prod.

### DIFF
--- a/docker/run
+++ b/docker/run
@@ -10,6 +10,7 @@ CONFIG_DIR=/etc/hub
 # LOG_DIR also exists in logback.xml
 LOG_DIR=/mnt/log
 DEBUG_PORT=3333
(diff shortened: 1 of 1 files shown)


lkemmerer:
Not 100% sure I follow this. Will it be configured for all hubs that run via docker and the only part that would enable it is exposing/binding the appropriate port on the container? I'm having trouble reasoning through what'll happen if that environment variable isn't set.

dkleckler:
It will default to a bogus hostname that will prevent the RMI registry from binding in an accessible way.  No errors, you just won't be able to hit it from outside of the container.
To get this to work, you need to be sure to bind 8888:8888 and also pass the host's IP as that environment variable.

PR #1171, opened 2019-05-09
Fix issue

manjula-chikkanagappa:
I have fixed SpokeTtlEnforcer service registration issues.
We should clean up HubMain sometime. Its too cluttered and not comprehensible.


src/main/java/com/flightstats/hub/app/HubMain.java
```diff
@@ -169,14 +176,23 @@ public Server startServer() throws Exception {
...
                 StatsDReporterLifecycle.class)
                 .map(injector::getInstance)
                 .collect(Collectors.toList());
+
         if (storageBackend == StorageBackend.aws && !appProperties.isReadOnly()) {
             services.add(injector.getInstance(S3WriteQueueLifecycle.class));
+            if (spokeProperties.isTtlEnforced()) {
+                services.add(new SpokeTtlEnforcerService(SpokeStore.WRITE,
+                        injector.getInstance(SpokeTtlEnforcer.class)));
+                services.add(new SpokeTtlEnforcerService(SpokeStore.READ,
```
dkleckler: We'll want to be able to run the TTL enforcers in r/o mode, or we could run into the same inode issue for read-heavy batch workloads.

src/main/java/com/flightstats/hub/app/HubMain.java
```diff
@@ -63,11 +67,12 @@
 @Slf4j
 public class HubMain {
 
-    private static final DateTime startTime = new DateTime();
-    private static AppProperties appProperties = new AppProperties(PropertiesLoader.getInstance());
-    private static SystemProperties systemProperties = new SystemProperties(PropertiesLoader.getInstance());
-    private static ZookeeperProperties zookeeperProperties = new ZookeeperProperties(PropertiesLoader.getInstance());
-    private final StorageBackend storageBackend = StorageBackend.valueOf(appProperties.getHubType());
+    private static DateTime startTime = new DateTime();
```
lkemmerer: Is there a reason this went from `final` to not?

src/main/java/com/flightstats/hub/spoke/SpokeTtlEnforcer.java
```diff
@@ -26,37 +23,33 @@
...
-            HubServices.register(new SpokeTtlEnforcerService());
-        }
+        this.spokeProperties = spokeProperties;
     }
 
     private Consumer<ChannelConfig> handleCleanup(AtomicLong evictionCounter) {
         return channel -> {
             int itemsEvicted = 0;
-            String channelPath = storagePath + "/" + channel.getDisplayName();
+            final String channelPath = storagePath + "/" + channel.getDisplayName();
```
chriskessel: FWIW, final inside a method is pretty low-value. It indicates this variable can't change. If your methods are so long that it's hard to tell whether a local variable is mutable then that methods has deeper issues...

src/main/java/com/flightstats/hub/app/HubMain.java
```diff
@@ -169,14 +176,23 @@ public Server startServer() throws Exception {
...
+
+        final List<Service> services = Stream.of(
                 InfluxdbReporterLifecycle.class,
                 StatsDReporterLifecycle.class)
                 .map(injector::getInstance)
                 .collect(Collectors.toList());
+
         if (storageBackend == StorageBackend.aws) {
             services.add(injector.getInstance(S3WriteQueueLifecycle.class));
+            if (spokeProperties.isTtlEnforced()) {
```
lkemmerer: I definitely like initializing/adding the lifecycle-type services here vs in the initializer. 👍 Thanks for moving it.

src/main/java/com/flightstats/hub/spoke/SpokeTtlEnforcer.java
```diff
@@ -26,37 +23,33 @@
...
         this.channelService = channelService;
         this.spokeContentDao = spokeContentDao;
         this.statsdReporter = statsdReporter;
-        this.storagePath = spokeProperties.getPath(spokeStore);
-        this.ttlMinutes = spokeProperties.getTtlMinutes(spokeStore) + 1;
-        if (spokeProperties.isTtlEnforced()) {
-            HubServices.register(new SpokeTtlEnforcerService());
-        }
+        this.spokeProperties = spokeProperties;
     }
```
dkleckler: Happy to see this service-registration-in-constructors model getting killed off.  :)

src/main/java/com/flightstats/hub/app/HubMain.java
manjula-chikkanagappa: nope. its an oversight

src/main/java/com/flightstats/hub/spoke/SpokeContentDao.java
```diff
@@ -83,11 +83,11 @@ public SpokeContentDao(Commander commander, SpokeProperties spokeProperties) {
         String command = String.format(GET_OLDEST_ITEM_COMMAND, storePath);
         int waitTimeSeconds = 3;
         String result = StringUtils.chomp(commander.runInBash(command, waitTimeSeconds));
-        if (StringUtils.isEmpty(result)) return Optional.absent();
```
lkemmerer: 👍

src/main/java/com/flightstats/hub/spoke/SpokeTtlEnforcer.java
```diff
@@ -27,38 +24,32 @@
...
 
     private Consumer<ChannelConfig> handleCleanup(AtomicLong evictionCounter) {
         return channel -> {
             int itemsEvicted = 0;
-            String channelPath = storagePath + "/" + channel.getDisplayName();
+            final String channelPath = storagePath + "/" + channel.getDisplayName();
             if (channel.isLive()) {
-                DateTime ttlDateTime = TimeUtil.stable().minusMinutes(ttlMinutes + 1);
+                final DateTime ttlDateTime = TimeUtil.stable().minusMinutes(ttlMinutes + 1);
```
lkemmerer: I'm not sure I follow why we're not setting ttl minutes in the initializer anymore. It means we could technically call this method before `initSpokeStore` is called. Generally I try to avoid any mutable attributes at the class instance level because of potential bugs like that...instead I'd rather pass those values as arguments if you don't know the values at initialization time.

lkemmerer: ...but in this case, you could probably add private getters for some of these values, since this is injecting `SpokeProperties`.

src/main/java/com/flightstats/hub/spoke/SpokeTtlEnforcer.java
```diff
@@ -74,46 +65,47 @@ public SpokeTtlEnforcer(SpokeStore spokeStore,
...
+    void cleanup(SpokeStore spokeStore) {
+        try {
+            initSpokeStore(spokeStore);
 
-    }
+            final long start = System.currentTimeMillis();
+            final AtomicLong evictionCounter = new AtomicLong(0);
 
+            log.info("running ttl cleanup");
+            TtlEnforcer.enforce(storagePath, channelService, handleCleanup(evictionCounter));
```
lkemmerer: Making a note to myself that we should try to get rid of this static at some point...

src/main/java/com/flightstats/hub/spoke/SpokeTtlEnforcer.java
manjula-chikkanagappa: SpokeTtlEnforcer needs to know SpokeStore  type to function. This comes from SpokeTtlEnforcerService at runtime. The spoke properties cannot be initialized in constructor since we spoekstore type is known only at runtime.  I would not worry about this since all the methods in this class is private except for cleanup method which sets spokestore type.

src/main/java/com/flightstats/hub/spoke/SpokeTtlEnforcer.java
```diff
@@ -74,46 +65,47 @@ public SpokeTtlEnforcer(SpokeStore spokeStore,
...
-        @Override
-        protected Scheduler scheduler() {
-            return Scheduler.newFixedRateSchedule(1, 1, TimeUnit.MINUTES);
-        }
+    void cleanup(SpokeStore spokeStore) {
+        try {
+            initSpokeStore(spokeStore);
 
-    }
+            final long start = System.currentTimeMillis();
```
lkemmerer: How do you feel about moving the execution time logging and exception handling out to the abstract scheduled service? That way the only responsibility of this method would be to clean up and log how many items were evicted. If you agree, you don't necessarily need to change it now..

src/main/java/com/flightstats/hub/spoke/SpokeTtlEnforcerService.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.google.common.util.concurrent.AbstractScheduledService;
+import lombok.extern.slf4j.Slf4j;
+
+import javax.inject.Inject;
+import java.util.concurrent.TimeUnit;
+
+@Slf4j
+public class SpokeTtlEnforcerService extends AbstractScheduledService {
+
```
Paul-Hess: Thanks for abstracting this out from a nested class. Most recently when this has been done, classes patterned like this that extend Abstract*Service we have used Lifecycle as a naming scheme that is a little more specifically semantic in this context. I think this should be renamed SpokeTtlEnforcerLifecycle for consistency

src/main/java/com/flightstats/hub/app/HubMain.java
```diff
@@ -169,14 +176,23 @@ public Server startServer() throws Exception {
     }
 
     private List<Service> getBeforeHealthCheckServices(Injector injector) {
-        List<Service> services = Stream.of(
+
```
chriskessel: This is super nit-picky, totally up to your group on style, by why do you add an extra blank link at the start of some methods, but not others?

manjula-chikkanagappa: Good catch :) I need to be consistent with this. When I think method is cluttered and not readable, I tend to group lines with blank line in between.

src/main/java/com/flightstats/hub/spoke/SpokeTtlEnforcer.java
lkemmerer: I'd probably opt to pass spokeStore around to methods that need it in that case. I've seen enough bugs with mutable state that I'll avoid it whenever I can.

manjula-chikkanagappa:
> This looks good. I'd probably get rid of the mutable instance variable in the spoke ttl enforcer and non-finalize variables in methods but those things are stylistic/conventions that I follow that we haven't come to a consensus on.

The variables we are talking about in the SpokeTtlEnforcer class are are object scoped and I prefer initializing as instance variables. The variables are encapsulated in the class.

PR #1172, opened 2019-05-10
Fix channel config parse error

dkleckler:
Fix channel config parse error and pull out the dynamo lifecycle services.


src/main/java/com/flightstats/hub/app/HubMain.java
```diff
@@ -176,15 +177,18 @@ public Server startServer() throws Exception {
...
-        if (storageBackend == StorageBackend.aws && !appProperties.isReadOnly()) {
-            services.add(injector.getInstance(S3WriteQueueLifecycle.class));
+        if (storageBackend == StorageBackend.aws) {
+            if (!appProperties.isReadOnly()) {
+                services.add(injector.getInstance(S3WriteQueueLifecycle.class));
+            }
+            services.addAll(createInstanceList(injector,
+                    DynamoChannelConfigDaoLifecycle.class,
+                    DynamoWebhookDaoLifecycle.class));
         }
```
manjula-chikkanagappa: Nice. Eventually this whole service registration thing should be in its own class.

src/test/java/com/flightstats/hub/dao/aws/DynamoChannelConfigDaoTest.java
```diff
@@ -35,4 +69,54 @@ void testSimple() {
...
+        Map<String,AttributeValue> channel = new HashMap<>();
+        channel.put("key", new AttributeValue(key));
+        channel.put("displayName", new AttributeValue(displayName));
+        channel.put("date", new AttributeValue().withN(String.valueOf(date)));
+        return channel;
+    }
+
+    private Map<String,AttributeValue> createBogusEntry() {
+        Map<String,AttributeValue> bogusRecord = new HashMap<>();
+        bogusRecord.put("i exist", new AttributeValue("to break your code"));
```
manjula-chikkanagappa: does this work? Spaces in key does not look right.

dkleckler: Yep.  At the end of the day, it's just a string.

src/main/java/com/flightstats/hub/app/HubMain.java
dkleckler: These particular ones may even die off.  I don't really think Hub should be responsible for configuring its own AWS resources.

src/test/java/com/flightstats/hub/dao/aws/DynamoChannelConfigDaoTest.java
```diff
@@ -1,27 +1,61 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
+@ExtendWith(MockitoExtension.class)
 class DynamoChannelConfigDaoTest {
 
     private static final Logger logger = LoggerFactory.getLogger(DynamoChannelConfigDaoTest.class);
     private static DynamoChannelConfigDao channelConfigDao;
+    @Mock private AmazonDynamoDB dbClient;
+    @Mock private DynamoProperties dynamoProperties;
```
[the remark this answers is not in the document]
dkleckler: Do we like that better?  I kinda prefer it on a single line so I can clearly see which of my ivars aren't real.

src/main/java/com/flightstats/hub/config/DynamoProperties.java
```diff
@@ -5,10 +5,28 @@
 public class DynamoProperties {
 
     private final PropertiesLoader propertiesLoader;
+    private final AppProperties appProperties;
 
     @Inject
-    public DynamoProperties(PropertiesLoader propertiesLoader) {
+    public DynamoProperties(PropertiesLoader propertiesLoader, AppProperties appProperties) {
```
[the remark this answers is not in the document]
dkleckler: Yeah.  Agreed it feels dirty.  _Can_ we kill off the legacy table names?

src/test/java/com/flightstats/hub/dao/aws/DynamoChannelConfigDaoTest.java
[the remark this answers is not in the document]
crozierm: (Historically) I think we mostly like IntelliJ auto-format and not having to configuring it, more than the actual format.

src/main/java/com/flightstats/hub/dao/aws/DynamoChannelConfigDao.java
```diff
@@ -84,112 +79,89 @@ public void upsert(ChannelConfig config) {
...
-            builder.mutableTime(new DateTime(Long.parseLong(item.get("mutableTime").getN()), DateTimeZone.UTC));
+    @VisibleForTesting
+    Optional<ChannelConfig> mapItem(Map<String, AttributeValue> item) {
+        ChannelConfig cfg = null;
+        try {
+            ChannelConfig.ChannelConfigBuilder builder = ChannelConfig.builder()
+                    .creationDate(new Date(Long.parseLong(item.get("date").getN())))
+                    .name(item.get("key").getS())
+                    .displayName(item.get("displayName").getS());
+            if (item.get("ttlDays") != null) {
```
chriskessel: It'd be nice to find a way to avoid all the if statments with a null object pattern or helper. Something like:
```
builder.ttlDays(item.getOrDefault("ttlDays"), AttributeValue.ALL_NULLS).getN());
builder.keepForever(item.getOrDefault("keepForever"), AttributeValue.ALL_NULLS).getBOOL());
builder.description(item.getOrDefault("ttlDays"), AttributeValue.ALL_NULLS).getS());
```

src/main/java/com/flightstats/hub/dao/aws/DynamoChannelConfigDao.java
```diff
@@ -204,32 +176,16 @@ private ChannelConfig mapItem(Map<String, AttributeValue> item) {
     }
 
     private void mapItems(List<ChannelConfig> configurations, ScanResult result) {
-        for (Map<String, AttributeValue> item : result.getItems()) {
-            configurations.add(mapItem(item));
-        }
+        result.getItems().forEach(
```
chriskessel: Also an option:
```
configurations.addAll(result.getItems().stream()
    .map(item -> mapItem(item))
```

src/main/java/com/flightstats/hub/dao/aws/DynamoChannelConfigDaoLifecycle.java
```diff
@@ -0,0 +1,71 @@
...
+            if (appProperties.isReadOnly()) {
+                String msg = String.format("Probably fatal error. Dynamo channel config table doesn't exist for r/o node.  %s", tableName);
+                log.error(msg);
+                throw new IllegalArgumentException(msg);
+            } else {
+                createTable(tableName, throughput);
+            }
+        } else if (!appProperties.isReadOnly()){
+            dynamoUtils.updateTable(tableName, throughput);
+        }
```
chriskessel: FWIW, an if/elseif without a trailing else tends to make me uncomfortable. There's an implied set of conditions that allows a fall through and I have to re-figure out what that is when I read the code. I tend to add an else that logs the reason or at least a comment.
```
else {
    log.info("Read only hub and we validated the dynamoTable already exists.");
}
```
or
```
// else - Read only hub and we validated the dynamoTable already exists
```

src/main/java/com/flightstats/hub/dao/aws/DynamoWebhookDaoLifecycle.java
```diff
@@ -0,0 +1,46 @@
...
+    private void initialize() {
+        if (appProperties.isReadOnly()) {
+            if (!dynamoUtils.doesTableExist(dynamoProperties.getWebhookConfigTableName())) {
+                String msg = String.format(
+                        "Probably fatal error. Dynamo webhook config table doesn't exist for r/o node.  %s",
+                        dynamoProperties.getWebhookConfigTableName());
+                log.error(msg);
+                throw new IllegalArgumentException(msg);
+            }
+            return;
```
chriskessel: Personal style, but I'd use an else and remove this return to make it explicit that the two things are mutually exclusive situations. It can also avoid subtle bugs. 
Buggy code:
```
if (thing_is_true) {
   do thing;
   // oops...they coder forgot to put a return here and it ends up always executing the statement below!
}
do_mutually_exclusive_thing;
```
Avoids even the possibility of that bug:
```
if (thing_is_true) {
    do thing;
}  else {
    do mutually exclusive thing;
}
```

src/test/java/com/flightstats/hub/dao/aws/DynamoChannelConfigDaoTest.java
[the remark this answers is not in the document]
chriskessel: I'd go with whatever IntelliJ auto-formats to. In the Dev Services code base, I habitually auto-reformat the file every time I finish editing some code so any code out of conformance ends up "fixed" anyway.

For the record, I'm with dkleckler, I like the one-liner for this, but it's worth it to stick to the norms if it means never worrying about auto-formatting screwing up some hand-format.

Of course, you can decide as a group what IntelliJ settings to override, just make sure everyone has the same settings (and it might not be possible to configure IntelliJ for granular exceptions, like in this case).

src/main/java/com/flightstats/hub/config/DynamoProperties.java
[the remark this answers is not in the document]
lkemmerer: We'd either have to figure out how to do a no-downtime migration or take the hub down in order to rename tables. I feel like there are other more important pieces of tech debt that wouldn't be as impactful to deploy, but it wouldn't be a bad idea to track the issue if we're not already.

src/main/java/com/flightstats/hub/dao/aws/DynamoChannelConfigDao.java
lkemmerer: Another way I've seen is to done is something like 
```
Optional.ofNullable(item.get("blah"))
    .map(AttributeValue::getWhatever)
    .ifPresent(builder::blah)
```

src/main/java/com/flightstats/hub/dao/aws/DynamoChannelConfigDaoLifecycle.java
lkemmerer: I think this might be more readable if the blocks were separated as readonly/not-readonly.  So something like:
```
if (isReadOnly) {
  if (tableExists) {
    throwTheError;
  }
} else {
  if (tableExists) {
    updateTable();
  } else {
    createTable();
  }
}
```

src/main/java/com/flightstats/hub/config/DynamoProperties.java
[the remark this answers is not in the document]
manjula-chikkanagappa: Created a ticket. https://jira.flightglobal.io/browse/PDXDDT-820

src/main/java/com/flightstats/hub/dao/aws/DynamoWebhookDaoLifecycle.java
dkleckler: Yep, done.

src/test/java/com/flightstats/hub/dao/aws/DynamoChannelConfigDaoTest.java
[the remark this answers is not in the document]
dkleckler: Switched.

src/main/java/com/flightstats/hub/dao/aws/DynamoChannelConfigDaoLifecycle.java
dkleckler: Yep.  Fixed.

PR #1173, opened 2019-05-13
Parallel tests

Paul-Hess:
this seems like an easy win, was able to get it running while debugging other stuff in the background. For general unit testing it halves the total run time 👍


lkemmerer:
Sweet! Does it pass in CI and everything?

src/test/resources/junit-platform.properties
```diff
@@ -0,0 +1,2 @@
+junit.jupiter.execution.parallel.enabled=TRUE
```
Paul-Hess: This makes it to run tests in parallel on N threads
https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution

src/test/java/com/flightstats/hub/webhook/WebhookStateReaperTest.java
```diff
@@ -30,6 +32,7 @@
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+@Execution(ExecutionMode.SAME_THREAD)
```
Paul-Hess: A few of our tests run over each other's state if run in parallel, i.e. trying to create the same node in zookeeper at the same time as another. This execution mode overrides the default parallel run setting.

Paul-Hess:
About to run it in Jenkins, have only passed it locally so far

Paul-Hess:
> Sweet! Does it pass in CI and everything?

>> About to run it in Jenkins, have only passed it locally so far

Passed and shaved 1+ minute off the average build time

src/test/java/com/flightstats/hub/webhook/WebhookStateReaperTest.java
dkleckler: Cool.  I've definitely run across some classes that can't be internally parallelized.  Nice feature.

src/test/resources/junit-platform.properties
dkleckler: Nice. This way we have to explicitly flag our tests that require SAME_THREAD...which probably marks them as good refactoring opportunities.

PR #1175, opened 2019-05-14
Remove circular dependency

manjula-chikkanagappa:
1) Removed the circular dependency between ChannelService and ReplicationManager
2) TagWebhook and WebhookService are tightly coupled and could not break circular dependency. This requires some greater refactor and is out of scope of the ticket. I have used field injection for Tagwebhook field for now instead of constructor injection.
3) Merged Damon’s read only hub code changes
4) Cleaned up static injection through our hub. HubProvider.class is removed.
5) I have used Constants class as for constants used thought our hub.  This approach is better than making fields public/package access within class.
6) Created Guice child injector for dependency injection to work in web socket  resources.


lkemmerer:
For a second I was surprised to see that there were 50k+ new lines of code, but it looks like there are a few really large test log output files that were inadvertently added to the PR.

src/main/java/com/flightstats/hub/config/binding/ClusterHubBindings.java
```diff
@@ -63,7 +63,7 @@ protected void configure() {
         bind(AppUrlCheck.class).asEagerSingleton();
 
         bind(SpokeDecommissionManager.class).asEagerSingleton();
-        bind(SpokeTtlEnforcer.class).asEagerSingleton();
+        bind(SpokeTtlEnforcer.class);
```
dkleckler: Need to pull the @Singleton off the class itself as well.

src/test/java/com/flightstats/hub/webhook/WebhookTest.java
```diff
@@ -1,32 +1,38 @@
...
 class WebhookTest {
 
     private static final int CALLBACK_TIMEOUT_DEFAULT_IN_SEC = 120;
+    @Mock
+    private ContentRetriever contentRetriever;
     private Webhook webhook;
 
     @BeforeEach
     void setUp() {
+        MockitoAnnotations.initMocks(this);
```
lkemmerer: Does this do the same thing as adding an annotation to the class? If so, is there an advantage to doing it one way or the other?

src/test/javascript/intout
```diff
@@ -0,0 +1,6809 @@
+
```
lkemmerer: I think  this file can be removed.

src/test/javascript/intout1
```diff
@@ -0,0 +1,6899 @@
+
```
lkemmerer: This file can go, too.

src/main/java/com/flightstats/hub/webhook/WebhookService.java
```diff
@@ -37,27 +46,33 @@ public WebhookService(@Named("Webhook") Dao<Webhook> webhookDao,
...
+            this.upsertTagWebhook(webhook);
+            return preExisting;
         }
         if (preExisting.isPresent()) {
-            Webhook existing = preExisting.get();
-            ContentPath newStartingKey = webhook.getStartingKey();
-            if (newStartingKey != null && newStartingKey.getTime().compareTo(new DateTime(1980, 1, 1, 1, 1)) >= 0
+            final Webhook existing = preExisting.get();
+            final ContentPath newStartingKey = webhook.getStartingKey();
+            if (newStartingKey != null
```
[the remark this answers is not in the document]
manjula-chikkanagappa: As discussed,  this is out of scope of ticket. Since this particular ticket is about removing static injection and  it touches most parts of the hub, I would avoid any other refactor here. This will help in debugging any issues that might crop up. Any other code refactor should be taken up separately and should be refined to particular functional area.

src/main/java/com/flightstats/hub/app/ShutdownManager.java
```diff
@@ -38,25 +46,25 @@ public boolean shutdown(boolean useLock) throws Exception {
         statsdReporter.event("Hub Restart Shutdown", "shutting down", tags);
         statsdReporter.mute();
 
-        HubHealthCheck healthCheck = HubProvider.getInstance(HubHealthCheck.class);
-        if (healthCheck.isShuttingDown()) {
+        if (this.hubHealthCheck.isShuttingDown()) {
```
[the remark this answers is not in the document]
manjula-chikkanagappa: Not sure of that. Is this convention documented somewhere? If there is a local variable/argument by the same name, instance variable will be ignored. "this" keyword ensures that use of instance variables.

src/main/java/com/flightstats/hub/util/Constants.java
```diff
@@ -0,0 +1,33 @@
+package com.flightstats.hub.util;
+
+public class Constants {
```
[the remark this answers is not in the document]
manjula-chikkanagappa: I have put these constants in separate files. Naming these classes is real challenge.

src/main/java/com/flightstats/hub/app/HubMain.java
```diff
@@ -121,12 +125,12 @@ private void startZookeeperIfSingle() {
     public Server startServer() throws Exception {
 
         List<AbstractModule> guiceModules = buildGuiceModules();
-        Injector injector = Guice.createInjector(guiceModules);
+        injector = Guice.createInjector(guiceModules);
```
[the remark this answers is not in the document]
manjula-chikkanagappa: We discussed this. This will come in as part of HubMain refactor.

src/main/java/com/flightstats/hub/app/ShutdownManager.java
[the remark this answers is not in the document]
lkemmerer: You're right that local variables with the same name will cause the instance variables to be shadowed, but I think that kind of duplicate naming is confusing and should be avoided where possible. The only exception to that case is in constructors where the main (and hopefully only) purpose is to map arguments/locals to the instance variables. I usually opt to use clear, distinct names and drop `this` as I find it makes the code a little noisy and less readable.

We don't have this stylistic choice documented anywhere; it's been a common convention on the trips teams and seems to generally be the case in the hub, iirc. This is another decision that we should get consensus on so that code we touch is consistent. @Paul-Hess and @dkleckler?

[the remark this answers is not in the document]
chriskessel: The only place using ```this``` tends to be important is in the constructor because of shadowing. The constructor is about the only place where a shadowing variable of the same name is even reasonable. Anywhere else, there shouldn't be any reason to have a local variable of the same name as an instance variable (and if you did, just pick a different name). 

Seeing an extraneous ```this``` then makes me think the clarification exists because of a conflict and I end up confusedly looking for where the conflicting variable is defined.

IntelliJ does a good job of colorization so there's little chance of confusion when you're working in the code.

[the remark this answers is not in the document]
manjula-chikkanagappa: I did a quick sweep on hub code and git rid of redundant this and final keywords to maintain consistency across the app.

src/main/java/com/flightstats/hub/channel/ChannelEarliestResource.java
```diff
@@ -1,13 +1,21 @@
 package com.flightstats.hub.channel;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.flightstats.hub.app.HubProvider;
 import com.flightstats.hub.dao.ChannelService;
-import com.flightstats.hub.model.*;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import com.flightstats.hub.dao.aws.ContentRetriever;
```
Paul-Hess: Is this ```import com.flightstats.hub.dao.ChannelService;``` still necessary, seems unused?

src/main/java/com/flightstats/hub/channel/ChannelLatestResource.java
```diff
@@ -1,12 +1,22 @@
 package com.flightstats.hub.channel;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.flightstats.hub.app.HubProvider;
 import com.flightstats.hub.dao.ChannelService;
-import com.flightstats.hub.model.*;
+import com.flightstats.hub.dao.aws.ContentRetriever;
```
Paul-Hess: Same here

src/main/java/com/flightstats/hub/channel/TagContentResource.java
```diff
@@ -196,32 +165,30 @@ public Response getValue(@PathParam("tag") String tag,
                              @PathParam("s") int second,
                              @PathParam("ms") int millis,
                              @PathParam("hash") String hash,
-                             @HeaderParam("Accept") String accept
-    ) {
-        long start = System.currentTimeMillis();
-        ContentKey key = new ContentKey(year, month, day, hour, minute, second, millis, hash);
-        ItemRequest itemRequest = ItemRequest.builder()
+                             @HeaderParam("Accept") String accept) {
+        final ContentKey key = new ContentKey(year, month, day, hour, minute, second, millis, hash);
```
Paul-Hess: I think we agreed that we should not be adding final to method scoped variables

manjula-chikkanagappa: I cleaned up some more. We will clean up rest as we refactor.

src/main/java/com/flightstats/hub/channel/ChannelLatestResource.java
manjula-chikkanagappa: Yes ChannelService is still used.

src/main/java/com/flightstats/hub/channel/ChannelEarliestResource.java
manjula-chikkanagappa: Yes ChannelService is still used.

src/test/java/com/flightstats/hub/channel/ChannelValidatorTest.java
```diff
@@ -10,29 +10,34 @@
...
+    private Dao<ChannelConfig> channelConfigDao;
     private ChannelValidator validator;
 
     @BeforeEach
     void setUp() {
-        channelService = mock(ChannelService.class);
-        validator = new ChannelValidator(channelService);
-        when(channelService.channelExists(any(String.class))).thenReturn(false);
+        initMocks(this);
+        validator = new ChannelValidator(channelConfigDao);
```
dkleckler: I'd change this to @ExtendWith the MockitoExtension.

manjula-chikkanagappa: Yes I agree with you. ClusterContentService and ChannelService requires some serious refactoring. I realized it was too much work in the current ticket and out of scope for this ticket. We should definitely address it in another ticket.

PR #1176, opened 2019-05-15
jmx connection issue fix

manjula-chikkanagappa:


docker/run
```diff
@@ -40,7 +40,7 @@ JAVA_OPTS="
  -Dsun.rmi.dgc.server.gcInterval=300000
  -agentlib:jdwp=transport=dt_socket,address=$DEBUG_PORT,server=y,suspend=n
  -Djboss.platform.mbeanserver
- -Dcom.sun.management.jmxremote
+ -Dcom.sun.management.jmxremote=true
```
dkleckler: Hmm...this shouldn't be necessary, since the prop defaults to true.  Can add it for clarity though.  I can connect to my local docker without this.

manjula-chikkanagappa: ok. I had trouble connecting locally until I set this property to true. But I am able to connect to iad .dev remotely without this setting. I agree with you that its better to set this property explicitly for clarity.

PR #1177, opened 2019-05-15
change spokettlenforcer bean type to prototype

manjula-chikkanagappa:


src/main/java/com/flightstats/hub/config/binding/ClusterHubBindings.java
```diff
@@ -63,7 +63,7 @@ protected void configure() {
         bind(AppUrlCheck.class).asEagerSingleton();
 
         bind(SpokeDecommissionManager.class).asEagerSingleton();
-        bind(SpokeTtlEnforcer.class).asEagerSingleton();
+        bind(SpokeTtlEnforcer.class);
```
lkemmerer: I think we can use implicit/JIT binding for this class unless we start using one for read and one for write (in which case I think we could bind them as eager singletons).

src/main/java/com/flightstats/hub/spoke/SpokeTtlEnforcer.java
```diff
@@ -20,7 +19,6 @@
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-@Singleton
```
lkemmerer: I'm going to bring up my dislike of having state in this class again. When this was a singleton, the lifecycle runs were probably stomping on the spokeStore value; if we'd had two instances (one for read and one for write) or passed the spokeStore variable around as an argument, this particular bug may not have occurred.

I think we should have a team conversation about what conventions we'd like to adopt for mutability and the use of `final`s. We've got some differing opinions and I'd like to come to a consensus to make code reviews a little easier going forward.

chriskessel: Also, if it's mutable and used across threads, the non-final variables should be volatile, protected by synchronized, or the entire SpokeTtlEnforcer safely published via access through a java.util.concurrent item (e.g. AtomicReference, ConcurrentHashMap, or some such).

lkemmerer: @chriskessel That paragraph of options and things to remember is why I try to make everything immutable. :P

manjula-chikkanagappa: Currently, SpokeTtlEnforcer is used in single threaded mode. We do not have to worry about concurrency/synchronization..

lkemmerer: Prior to the un-Singleton-ization, it was being used by two separate services (read and write spoke ttl lifecycles), very possibly in two separate threads, though, right?
I would argue that we should try to avoid writing code that needs to be changed in order to work properly in a multithreaded environment.

manjula-chikkanagappa: > Prior to the un-Singleton-ization, it was being used by two separate services (read and write spoke ttl lifecycles), very possibly in two separate threads, though, right?
> I would argue that we should try to avoid writing code that needs to be changed in order to work properly in a multithreaded environment.

Programming for multithreaded environment is no trivial task. Especially in this case, we know that the functionality will be executed in single thread.  We do not want to over optimize it. On the second thought, it is not even optimization. We are adding unnecessary complexity to the program. I believe in and follow YAGNI and KISS programming principles.

lkemmerer: In my experience, removing state from classes/making things immutable reduces complexity and makes behavior more deterministic and easier to reason about. We should get together and talk this over as a team; I think getting everybody's input will be helpful.

dkleckler: For my 2c, I'd make these three vars final, bind SpokeStore.WRITE and READ variants of SpokeTTLEnforcer, and have the lifecycles grab their associated version.  Some classes can't avoid mutable state and it should be clear they can't be shared.  In this case, we init the state to "fixed" values every iteration of the lifecycle, so we really intend to have two functionally distinct versions of this anyway.

manjula-chikkanagappa: I made spokettlenforcer class immutable. Earlier i had thought multiple times, but did not find a way of doing it. May be too much refactor fatigue.
Also, originally spokeTtlenforcer was bound in ClusterHubBindings class. Not sure why. I have currently put it in HubBindings class. As per my understanding, SpokeTtlEnforcer runs both in single and cluster mode.

lkemmerer: SpokeTtlEnforcer should only run in clustered/AWS mode (that's what it's doing in the code that's on production, anyway). I think Paul's working on a fix for that, but you might want to check with him.

manjula-chikkanagappa: Ok. SpokeTtlEnforcerService is enabled only in cluster mode.

PR #1178, opened 2019-05-16
ttl enforcer refactor/test

Paul-Hess:


src/main/java/com/flightstats/hub/dao/TtlEnforcer.java
```diff
@@ -2,38 +2,46 @@
...
+            Set<String> channelSet = channels.stream()
+                    .map(ChannelConfig::getLowerCaseName)
+                    .collect(Collectors.toSet());
+            Stream.of(pathArray)
+                    .map(String::toLowerCase)
+                    .filter(channel -> !channelSet.contains(channel.toLowerCase()) && !channel.equals("lost+found"))
+                    .forEach(dir -> {
+                        String dirPath = path + "/" + dir;
+                        log.info("removing dir without channel {}", dirPath);
+                        commander.runInBash("rm " +  " -rf " +  dirPath, 1);
```
Paul-Hess: This calls it's own static method, which I would like to fix but Commander is used all over the place and it would be a much more involved PR to get into that

src/main/java/com/flightstats/hub/dao/TtlEnforcer.java
```diff
@@ -2,38 +2,47 @@
...
+    private static final String LOST_AND_FOUND_DIR = "lost+found";
+    private final Commander commander;
+
+    @Inject
+    public TtlEnforcer(Commander commander) {
+        this.commander = commander;
+    }
+
+    public void enforce(String path, ChannelService channelService,
+                        Consumer<ChannelConfig> channelConsumer) {
```
lkemmerer: While we're in here, can we rename this to something more specific? We pass methods called `handleCleanup` into both usages of this method, so that might inform some better name (`itemCleanupHandler`?). I haven't looked closer to see if there's anything in those consumers that's common to both...

src/main/java/com/flightstats/hub/dao/TtlEnforcer.java
```diff
@@ -2,38 +2,47 @@
...
-                logger.info("removing dir without channel {}", dirPath);
-                Commander.run(new String[]{"rm", "-rf", dirPath}, 1);
-            }
+            String[] pathArray = Optional.ofNullable(new File(path).list()).orElse(new String[]{});
+            Collection<ChannelConfig> channels = channelService.getChannels();
+            channels.forEach(channelConsumer);
+            Set<String> channelSet = channels.stream()
+                    .map(ChannelConfig::getLowerCaseName)
+                    .collect(Collectors.toSet());
+            Stream.of(pathArray)
```
lkemmerer: Oh my, this is so much more readable. Thank you!

src/main/java/com/flightstats/hub/dao/TtlEnforcer.java
lkemmerer: I only mention looking for commonalities, because having this generic consumer (that's supposed to do the bulk of the enforcing but is kind of a black box) is a little strange. I wonder if we could make it a function that returns files to be deleted or something.

lkemmerer: ...that sort of decoupling of finding and removing files might have performance implications. Hm. Nevermind.

src/test/java/com/flightstats/hub/dao/TtlEnforcerTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @BeforeEach
+    void setup() {
+        initMocks(this);
+        ttlEnforcer = new TtlEnforcer(commander);
+    }
+
+    @Test
+    void enforce_deletesFiles_affirmThreeFilesDeleted(@TempDir File spoke) {
```
lkemmerer: I might rename this to make it explicit that it's deleting directories that don't match up with channels or LOST_AND_FOUND_DIR rather than just deleting files.

src/test/java/com/flightstats/hub/dao/TtlEnforcerTest.java
```diff
@@ -0,0 +1,103 @@
...
+        assertEquals(0, getSpokePathList(spoke).length);
+        assertTrue(createMockSpokeFiles(spoke, "a", "b", "c", LOST_AND_FOUND_DIR));
+        ttlEnforcer.enforce(spoke.getPath(), channelService, callback);
+        verify(commander, times(3)).runInBash(anyString(), anyInt());
+    }
+
+    @Test
+    void enforce_deletesUpperCaseFiles_affirmTwoFilesDeleted(@TempDir File spoke) {
```
lkemmerer: Same comment as above, I think.

PR #1180, opened 2019-05-20
Adjust some log levels to decrease some noise

lkemmerer:
I was trying to work through some test logs in graylog and there was a lot of noise. This adjusts a couple of the more egregious culprits on low-volume systems and moves some others up to warn or error depending on how recoverable or exceptional the situation is.


src/main/java/com/flightstats/hub/cluster/DistributedLeaderLockManager.java
```diff
@@ -41,10 +41,10 @@ public boolean tryAcquireLock(LeadershipLock leadershipLock, long waitTimeout, T
                 log.debug("acquired {} {}", leadershipLock.getLockPath(), leadership.hasLeadership());
                 return true;
             } else {
-                log.debug("unable to acquire {} ", leadershipLock.getLockPath());
+                log.warn("unable to acquire {} ", leadershipLock.getLockPath());
             }
         } catch (Exception e) {
-            log.warn("oh no! issue with " + leadershipLock.getLockPath(), e);
+            log.error("oh no! issue with " + leadershipLock.getLockPath(), e);
             release(leadershipLock);
```
manjula-chikkanagappa: you can parameterize these log statements instead of string concatenation. Look at the example on line 44. This applies to all other log statements.

PR #1184, opened 2019-05-22
Refactor channel latest svc

dkleckler:
Touched a lot of files, but it's mostly a rename, killing the latest cache update service, and using the ZK cache for latest content reads, regardless of whether the latest content is sitting on a spoke or in S3.


src/main/java/com/flightstats/hub/cluster/ClusterCacheDao.java
```diff
@@ -79,27 +78,20 @@ public ContentPath get(String name, ContentPath defaultPath, String basePath) {
...
-
-    public void updateDecrease(ContentPath nextPath, String name, String basePath) {
-        update(nextPath, name, basePath, (existing) -> nextPath.compareTo(existing.key) < 0);
+    public void setIfBefore(ContentPath nextPath, String name, String basePath) {
+        setPathValueIf(nextPath, name, basePath, (existing) -> nextPath.compareTo(existing.key) < 0);
     }
 
-    public void updateIncrease(ContentPath nextPath, String name, String basePath) {
-        update(nextPath, name, basePath, (existing) -> nextPath.compareTo(existing.key) > 0);
+    public void setIfAfter(ContentPath nextPath, String name, String basePath) {
```
chriskessel: I can't figure out what after means. After what? I don't see an integer or timestamp, so I'm not sure how to interpret 'after'. Maybe setIfNewer() or setIfMoreRecent() or some other name might be more descriptive?

If this is just me being ignorant of a term common to the hub, then ignore me.

src/main/java/com/flightstats/hub/cluster/ClusterCacheDao.java
```diff
@@ -163,20 +157,28 @@ public void delete(String name, String basePath) {
...
         byte[] bytes = curator.getData().storingStatIn(stat).forPath(path);
         Optional<ContentPath> pathOptional = ContentPath.fromUrl(new String(bytes, StandardCharsets.UTF_8));
-        return new LastUpdated(pathOptional.get(), stat.getVersion());
+        return new MostRecentData(pathOptional.get(), stat.getVersion());
     }
 
-    private class LastUpdated {
-        ContentPath key;
-        int version;
+    private class MostRecentData {
```
chriskessel: 'Data' is almost always a bad choice for a name in a class (or any name, barring something like raw bytes from a file). Everything is data. Clean code calls them "noise words", words that really don't tell you any more than if they were left off. It specifically calls out 'data' and 'info' as common culprits. 

MostRecentPathVersion? MostRecentKey?

src/main/java/com/flightstats/hub/cluster/ClusterCacheDao.java
```diff
@@ -163,20 +157,28 @@ public void delete(String name, String basePath) {
...
+        private ContentPath key;
+        private int version;
 
-        private LastUpdated(ContentPath key, int version) {
+        private MostRecentData(ContentPath key, int version) {
             this.key = key;
             this.version = version;
         }
+
+        public ContentPath getKey() {
```
chriskessel: Why not use lombok ```@Value```?

src/main/java/com/flightstats/hub/cluster/ClusterCacheDao.java
```diff
@@ -163,20 +157,28 @@ public void delete(String name, String basePath) {
...
         Optional<ContentPath> pathOptional = ContentPath.fromUrl(new String(bytes, StandardCharsets.UTF_8));
-        return new LastUpdated(pathOptional.get(), stat.getVersion());
+        return new MostRecentData(pathOptional.get(), stat.getVersion());
     }
 
-    private class LastUpdated {
-        ContentPath key;
-        int version;
+    private class MostRecentData {
+        private ContentPath key;
```
chriskessel: These can be final.

src/main/java/com/flightstats/hub/dao/aws/ClusterContentService.java
```diff
@@ -143,33 +149,66 @@ public ClusterContentService(
     }
 
     @Override
-    public ContentKey insert(String channelName, Content content) throws Exception {
+    public ContentKey insert(String channelName, Content content) {
+        final Content spokeContent = calculateIndexIfLarge(channelName, content);
```
chriskessel: This final isn't necessary even though it's used in the lambda. Java should be able to determine that it's effectively final and compile/run just fine.

That said, this is one example of a case where final in a method would be a reasonable group style decision. It's adding value by indicating it's being used in a lambda. In fact, as I type that sentence, I'm pondering if that's something I should start doing...

Regardless, be consistent, there are other variables in the PR used in lambdas that aren't final.

src/main/java/com/flightstats/hub/dao/aws/ClusterContentService.java
```diff
@@ -143,33 +149,66 @@ public ClusterContentService(
...
+        // after stable() seconds, we want to write the cache entry if the write succeeded
+        AtomicReference<ContentKey> ref = new AtomicReference<>();
+
+        ScheduledFuture future = zkCacheStateUpdateExecutor.schedule(() -> {
+            if (ref.get() != null) {
+                latestContentCache.setIfAfter(channelName, ref.get());
+            }
+        }, contentProperties.getStableSeconds() +1, TimeUnit.SECONDS);
+
+        SortedSet<ContentKey> keys;
```
chriskessel: Group style choice, but I try to keep variables scoped as tightly as possible. Defining keys outside of the try scope isn't necessary.
```
try {
    SortedSet<ContentKey> keys = supplier.get();
    ref.set(keys.last());
    return keys;
} catch Exception (e) {
    future.cancel(true);
    throw e;
}
```

src/main/java/com/flightstats/hub/dao/aws/ClusterContentService.java
```diff
@@ -344,47 +383,79 @@ private void getValues(String channelName, Consumer<Content> callback, ContentPa
...
-            return getLatestImmutable(query);
-        } else if (query.getEpoch().equals(Epoch.MUTABLE)) {
-            return ContentService.chooseLatest(queryDirection(query), query);
-        } else {
-            Optional<ContentKey> latestImmutable = getLatestImmutable(query);
-            if (latestImmutable.isPresent()) {
-                return latestImmutable;
-            }
-            return ContentService.chooseLatest(queryDirection(query), query);
+        switch (query.getEpoch()) {
```
chriskessel: Nice change from if/else to switch, :+1:

src/main/java/com/flightstats/hub/spoke/SpokeWriteContentDao.java
```diff
@@ -89,14 +90,15 @@ public Content get(String channelName, ContentKey key) {
         String path = getPath(channel, limitKey);
         log.trace("latest {} {}", channel, path);
         traces.add("SpokeWriteContentDao.latest", channel, path);
+        Optional<ContentKey> key;
```
chriskessel: Another case where you can safely decrease the scope of this variable. Though, it looks like you intentionally increased the scope?
```
try {
    Optional<ContentKey> key = chronoStore.getLatestFromCluster(channel, path, traces);
    traces.add("SpokeWriteContentDao.latest", key);
    return key;
} catch Exception (e) {
    log.warn("what happened? " + channel, e);
    return Optional.empty();
}
```

src/main/java/com/flightstats/hub/spoke/SpokeWriteContentDao.java
```diff
@@ -58,11 +58,12 @@ public ContentKey insert(String channelName, Content content) {
...
 
     @Override
-    public SortedSet<ContentKey> insert(BulkContent bulkContent) throws Exception {
-        return SpokeContentDao.insert(bulkContent, (baos) -> {
+    public SortedSet<ContentKey> insert(BulkContent bulkContent) {
+        SortedSet<ContentKey> keys = SpokeContentDao.insert(bulkContent, (baos) -> {
             String channel = bulkContent.getChannel();
             return clusterWriteSpoke.insertToWriteCluster(channel, baos.toByteArray(), "bulkKey", channel);
         });
+        return keys;
```
chriskessel: Minor nit as there's no harm here, but IntelliJ should warn you're setting and returning the variable immediately and offer to reduce this to a single line. If you're doing it in 2 lines so you can set a breakpoint on the return, IntelliJ's debugger has a feature where it'll show you the return value of any method you just exited, which basically accomplishes the same thing.

src/main/java/com/flightstats/hub/cluster/ClusterCacheDao.java
dkleckler: "Newer"/"older" would work as well.  There's sort of a general expectation in hub code that "paths" have an intrinsic time so "after" and "newer" are synonymous to me within that context.

src/main/java/com/flightstats/hub/cluster/ClusterCacheDao.java
dkleckler: Yeah, for sure.  VersionedKey, since that's what it is.  
Interesting that we never do anything with versions in hub though...wonder if we could just kill that...

src/main/java/com/flightstats/hub/spoke/SpokeWriteContentDao.java
dkleckler: Yep, for sure.  There was more code here at one point.

src/main/java/com/flightstats/hub/spoke/SpokeWriteContentDao.java
dkleckler: The flow wasn't explicit with that catch-all return that could only actually be hit if there was an exception.  Returning directly is easier to read here; will switch to that.

src/main/java/com/flightstats/hub/dao/aws/ClusterContentService.java
dkleckler: Yep.

src/main/java/com/flightstats/hub/dao/aws/ClusterContentService.java
dkleckler: Yep...and I think we've landed on preferring not to use finals in methods.

src/main/java/com/flightstats/hub/cluster/ClusterCacheDao.java
dkleckler: We're not using it much in the rest of the hub code.  Would be nice to start though.

chriskessel: I've probably told this story before, but apparently the Hub did use lombok mostly in the past. Then Ben (no longer here) had some oddity with his system that wouldn't handle lombok (in IntelliJ I think) that affected no one else and no one else could duplicate. Regardless, Ben and Greg ripped out all lombok. 

Ryan came back from PTO and went "Huh? Where'd all the lombok go?" and since he wasn't leader of the group he just rolled with it.

tl;dr - the reasons for not using Lombok in hub probably aren't valid anymore.

src/main/java/com/flightstats/hub/cluster/ClusterCacheDao.java
lkemmerer: My 2c is that "newer" is a little bit clearer in that it kind of connotes the idea of comparing two items. With the hub being so time-centric, it's a little confusing about which after we're talking about (stable time, order of items, ttl...) if I don't stop to think about it.

src/main/java/com/flightstats/hub/cluster/ClusterCacheDao.java
lkemmerer: The only versions I can think of that might be related to the hub are S3 versioning (which I don't think the hub knows anything about) and the actual hub version that's deployed, which data shouldn't care about and isn't an integer. If you know offhand, what is this one supposed to mean?

src/main/java/com/flightstats/hub/cluster/ClusterCacheDao.java
dkleckler: Sounds good.  "Newer"/"older" it is then.

src/main/java/com/flightstats/hub/cluster/ClusterCacheDao.java
lkemmerer: One of our hubkeeping notes is to move back to Lombok where possible/convenient because of that...whole...thing.

dkleckler: Heh.  That make a lot of sense with how the code reads, actually.  ;-)
I'm all for re-lombokification.

src/main/java/com/flightstats/hub/cluster/ClusterCacheDao.java
dkleckler: Hmm...well...that's an interesting read.  ZNodes have an inherent version.  ZK gives us a node at version 11 and expects us to pass that back in to say, "we're making an update relative to version 11".  That way if there's a version 12 ZK can reject the update with a KeeperException.BadVersionException.
We catch that and retry if our condition is still true (newer | older).  So...definitely want to keep that around.  :)

src/test/java/com/flightstats/hub/cluster/ClusterCacheDaoTest.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @BeforeEach
+    void setUp() {
+        clusterCacheDao = new ClusterCacheDao(curator, new AppProperties(PropertiesLoader.getInstance()));
+    }
+
+    @Test
+    void testLifeCycle() {
+        String name = "testLifeCycle";
```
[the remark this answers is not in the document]
dkleckler: Yeah.  This test class is pretty old.  Only the init() and delete() calls aren't tested by other Tests these days.  Fun fact:  the "delete" portion of the test relies on the fact that get() will re-create a deleted node.  Broke these out.

src/test/java/com/flightstats/hub/cluster/ClusterCacheDaoTest.java
```diff
@@ -0,0 +1,136 @@
...
+        clusterCacheDao.setIfNewer(nextPath, name, BASE_PATH);
+        assertEquals(nextPath, clusterCacheDao.get(name, new MinutePath(), BASE_PATH));
+
+        nextPath = new MinutePath(nextPath.getTime().plusMinutes(1));
+        clusterCacheDao.setIfNewer(nextPath, name, BASE_PATH);
+        assertEquals(nextPath, clusterCacheDao.get(name, new MinutePath(), BASE_PATH));
+    }
+
+    @Test
+    void testSetIfBefore() {
```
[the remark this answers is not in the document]
dkleckler: Yep, fixed.

src/main/java/com/flightstats/hub/dao/aws/ClusterContentService.java
```diff
@@ -461,16 +518,16 @@ public void notify(ChannelConfig newConfig, ChannelConfig oldConfig) {
...
         }
     }
 
-    private void handleMutableTimeChange(ChannelConfig newConfig, ChannelConfig oldConfig) {
-        final ContentPath latest = lastContentPath.get(newConfig.getDisplayName(), ContentKey.NONE, CHANNEL_LATEST_UPDATED);
-        log.info("handleMutableTimeChange {}", latest);
-        if (latest.equals(ContentKey.NONE)) {
-            final DirectionQuery query = DirectionQuery.builder()
+    private void setLastToNewlyImmutableContent(ChannelConfig newConfig, ChannelConfig oldConfig) {
+        ContentPath latestOrNoneIfEmpty = latestContentCache.getLatest(newConfig.getDisplayName(), ContentKey.NONE);
```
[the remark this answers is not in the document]
dkleckler: Added this.

src/main/java/com/flightstats/hub/dao/ChannelService.java
```diff
@@ -196,7 +196,7 @@ private void checkZeroBytes(Content content, ChannelConfig channelConfig) {
 
     public Collection<ContentKey> insert(BulkContent content) {
         final BulkContent bulkContent = content.withChannel(contentRetriever.getDisplayName(content.getChannel()));
```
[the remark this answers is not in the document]
dkleckler: done

src/main/java/com/flightstats/hub/dao/aws/ClusterContentService.java
```diff
@@ -143,33 +149,66 @@ public ClusterContentService(
...
-        if (isWriteable(channelName)) {
-            s3SingleWrite(channelName, key);
+        return spokeContent;
+    }
+
+    private SortedSet<ContentKey> setStableCache(String channelName, Supplier<SortedSet<ContentKey>> supplier) {
+        // after stable() seconds, we want to write the cache entry if the write succeeded
+        AtomicReference<ContentKey> ref = new AtomicReference<>();
+
+        ScheduledFuture future = zkCacheStateUpdateExecutor.schedule(() -> {
```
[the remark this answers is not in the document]
dkleckler: Yep.
It's an unbounded queue, so there shouldn't be a hard limit.  4 is the core size, but the implementation will scale up to MAX_INT threads as needed to unload the queue.  Overall this will reduce load from hub itself in /latest->*/internal/latest for spoke-resident latest items at the cost of two ZK ops per insert.  
We could try to coalesce events for channels like FAA feeds in iad.prod, but we'd have to play looser with the meaning of "stable" time.
TBH, I would love to change hub's concept of "stable" to "cluster acknowledged" or "majority acknowledged".  It would simplify code like this and the reads that self-select "stable" time limits, and, it'd be more robust.  That's a decent chunk of work though.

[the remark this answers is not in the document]
lkemmerer: I didn't think about what we gained from this change beyond getting rid of that scheduled task and the wonky ZK update side effect. That doesn't seem so bad.

I have a feeling that if/when we replace the storage implementation for the hub with something third party, we'll be able to stop worrying about stable time. Not knowing the timeline for that feels like punting on deciding whether we should put aside some time or a task to scope that work out.

PR #1186, opened 2019-05-23
Switch to slf4j for all remaining classes.

dkleckler:


src/main/java/com/flightstats/hub/util/AbstractExceptionMapper.java
```diff
@@ -1,22 +1,21 @@
...
 
 import com.flightstats.hub.metrics.ActiveTraces;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import lombok.extern.slf4j.Slf4j;
 
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 
+@Slf4j
```
chriskessel: Just a note on ```@Slf4j``` and abstract class, they can be rather annoying because the log line lists the class name of the abstract class rather than the implementing class. That can make it really tough to figure out which implementing class is actually doing the logging.

The best solution I've come up with so far is to remove ```@Slf4j``` on the abstract class, put it on the implementing class, and have the implementing class pass it's logger to the super class in the constructor.
```
protected AbstractExeptionMapper(Logger log) {
    this.log = log;
}
```

Might not matter here since it's an exception mapper, but I thought I'd mention it.

PR #1187, opened 2019-05-23
Ensure we request the s3batch dao using Named, so we actually get the right one.

dkleckler:


Paul-Hess:
My batch storage system test passes against this deployed on iad dev

PR #1189, opened 2019-05-24
code review comments

manjula-chikkanagappa:


src/main/java/com/flightstats/hub/config/server/JettyServer.java
```diff
@@ -123,23 +124,28 @@ private ResourceConfig buildResourceConfig(Injector injector) {
...
-            sslContextFactory.setKeyStorePassword(password);
+            sslContextFactory.setKeyStorePath(getKeyStorePath());
+            sslContextFactory.setKeyStorePassword(getKeyStorePassword());
         }
+
         return sslContextFactory;
     }
 
+    private String getKeyStorePath() {
+        final String keyStorePath = appProperties.getKeyStorePath() + HubHost.getLocalName() + ".jks";
```
chriskessel: Another local ```final``` crept in.

src/main/java/com/flightstats/hub/config/server/JettyServer.java
```diff
@@ -123,23 +124,28 @@ private ResourceConfig buildResourceConfig(Injector injector) {
...
-            log.info("using key store path: {}", keyStorePath);
-
             sslContextFactory = new SslContextFactory();
-            sslContextFactory.setKeyStorePath(keyStorePath);
-
-            String keyStorePasswordPath = appProperties.getKeyStorePasswordPath();
-            URL passwordUrl = new File(keyStorePasswordPath).toURI().toURL();
-            String password = Resources.readLines(passwordUrl, StandardCharsets.UTF_8).get(0);
-            sslContextFactory.setKeyStorePassword(password);
+            sslContextFactory.setKeyStorePath(getKeyStorePath());
```
lkemmerer: I like the helper methods! 
When I said "short-circuiting" in the other PR I meant using a guard clause. D'oh. The pattern always bounces out of the method immediately if you don't meet criteria applicable to the method. You do end up with two return statements, but the first one at most is an if/log/return and nothing else, ever.
Since that wasn't part of the refactoring here, I was wondering if you have strong feelings against using them. If so, why?

src/test/java/com/flightstats/hub/config/ClusterServicesRegistrationTest.java
```diff
@@ -0,0 +1,161 @@
...
+                s3WriteQueueLifecycle,
+                customMetricsLifecycle,
+                appProperties,
+                spokeProperties,
+                periodicMetricEmitterLifecycle);
+        HubServices.clear();
+    }
+
+    @Test
+    @Order(1)
```
lkemmerer: I think I'm being dumb.  It looks like we're resetting all the static state... Why does order matter for these tests?

src/main/java/com/flightstats/hub/config/server/JettyServer.java
lkemmerer: whoops, I didn't check my email so I missed the answer in the other PR. Thanks.

src/test/java/com/flightstats/hub/config/ClusterServicesRegistrationTest.java
manjula-chikkanagappa: By default, unit tests run concurrently. 
HubServices registration uses static map to store all services registered. Since it is static there is only one instance per class. All unit tests are writing and reading from the same map instance and needs to be executed one after the other.

lkemmerer: Oh! For some reason I thought that using the `ExecutionMode.SAME_THREAD` would effectively serialize these without needing to add ordering.

manjula-chikkanagappa: ExecutionMode.SAME_THREAD serializes tests at Testsuite level. @TestMethodOrder(MethodOrderer.OrderAnnotation.class) does it within Test class. There are a other ways of ordering tests (alphabetically by tests name..).

PR #1192, opened 2019-05-28
Switch to official Guice<->HK2 bridge.

dkleckler:


dkleckler:
> Looks good, afaict. I'd probably run it on dev and monitor logs for JIT binding errors/NPEs/other exceptions just in case the hub is doing something the library doesn't handle as expected.

It made it through all test tests out on dev a bit ago, so I think we're in good shape.  Bonus points that the bridge devs agree with us on what should happen when you request a non-Named binding when only Named bindings exist.
https://ddt-jenkins.pdx.prod.flightstats.io/job/run-hub-development-integration/324/

PR #1193, opened 2019-05-28
PDXDDT-800 Convert existing Config POJO + Provider pattern to use new…

manjula-chikkanagappa:
Convert existing Config POJO + Provider pattern to use new Properties class pattern.


src/main/java/com/flightstats/hub/metrics/InfluxdbReporterProvider.java
```diff
@@ -3,53 +3,65 @@
...
-            .tag("cluster", metricsConfig.getClusterTag())
-            .tag("host", metricsConfig.getHostTag())
-            .tag("version", metricsConfig.getAppVersion())
-            .build();
+                .protocol(protocol)
+                .convertRatesTo(TimeUnit.SECONDS)
+                .convertDurationsTo(TimeUnit.MILLISECONDS)
+                .filter(MetricFilter.ALL)
+                .skipIdleMetrics(false)
+                .tag("role", metricsProperties.getTagsRole())
```
Paul-Hess: I think that it might be a little bit more consistent to change these two names to getRoleTag, getTeamTag OR change getClusterTag to match these, I prefer the former personally

src/main/java/com/flightstats/hub/config/properties/DatadogMetricsProperties.java
```diff
@@ -11,7 +11,7 @@ public DatadogMetricsProperties(PropertiesLoader propertiesLoader) {
         this.propertiesLoader = propertiesLoader;
     }
 
-    public String getUrl() {
+    public String getApiUrl() {
```
chriskessel: If it's returning a URL would it make sense to define the return type as URL?

src/main/java/com/flightstats/hub/config/properties/MetricsProperties.java
```diff
@@ -0,0 +1,45 @@
...
+    public String getRoleTag() {
+        return propertiesLoader.getProperty("metrics.tags.role", "hub");
+    }
+
+    public String getTeamTag() {
+        return propertiesLoader.getProperty("metrics.tags.team", "development");
+    }
+
+    public String getHostTag() {
+        return HubHost.getLocalName();
```
chriskessel: Can you inject an instance of HubHost? Even if that instance is just HubHost.getInstance() provided by a ```@Provider``` it'd start the path of decoupling from the static methods. Though you'd have to define non-static verisons of methods like ```getLocalName()```.

manjula-chikkanagappa: yes i thought of that. Created a task for it.

PR #1194, opened 2019-05-28
Cleanup fileutil

manjula-chikkanagappa:
Move properties from Fileutil class to properties class.
Rename FileUtil to SpokeFile.


dkleckler:
Looks good.  Nice to not have a Util class that knows how to go find state...

PR #1196, opened 2019-05-29
Cleanup hubhost

manjula-chikkanagappa:
Convert all HubHost static methods to non static methods and rename methods 
Rename HubHost to LocalHostProperties
Move LocalHostProperties to config properties package


src/test/java/com/flightstats/hub/util/HubUtilsTest.java
```diff
@@ -8,11 +8,11 @@
...
-    private static final String HUT_TEST = "test_0_HubUtilsTest" + StringUtils.randomAlphaNumeric(6);
+   /* private static final String HUT_TEST = "test_0_HubUtilsTest" + StringUtils.randomAlphaNumeric(6);
     private static HubUtils hubUtils;
     private static String channelUrl;
 
-    /*@BeforeAll
+    @BeforeAll
     static void setUpClass() {
         hubUtils = IntegrationTestSetup.run().getInstance(HubUtils.class);
         channelUrl = create();
```
dkleckler: Should this be commented out?  Seems like we should fix the tests, delete them, or at-ignore them if we're going to fix them soon so we at least get compile-time checking.

manjula-chikkanagappa: I removed the comment. This test sometimes takes very long time to finish execution (i think it is due to timeout settings). So I generally comment it out. There is a ticket open to look into this issue.

dkleckler: Ah, yeah, I have that problem on my box.  Running the tests serially instead of in parallel is the work-around I've been using.

src/main/java/com/flightstats/hub/cluster/CuratorCluster.java
```diff
@@ -78,20 +82,19 @@ public void register() {
             log.info("this hub is read only, not registering");
             return;
         }
-        String host = Cluster.getHost(useName);
         try {
-            log.info("registering host {} {}", host, clusterPath);
-            curator.create().withMode(CreateMode.EPHEMERAL).forPath(getFullPath(), host.getBytes());
+            log.info("registering host {} {}", localHostProperties.getHost(useName), clusterPath);
```
lkemmerer: I think I'd've been tempted to pull a `getHost()` out into a private method since it's being used in this method 4 times and again in a few other methods.

src/main/java/com/flightstats/hub/spoke/SpokeManager.java
```diff
@@ -383,4 +391,13 @@ public void run() {
         return countDownLatch.await(60, TimeUnit.SECONDS);
     }
 
+    /**
+     * @return the localhost's server
+     */
+    private Collection<String> getLocalServer() {
+        List<String> server = new ArrayList<>();
+        server.add(localHostProperties.getHost(false));
+        return server;
```
lkemmerer: I think this could be done in one line with `return Arrays.asList(localHostProperties.getHost(false));`

src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
```diff
@@ -22,17 +21,15 @@
...
     @Inject
-    public InternalWebhookClient(@Named("HubCuratorCluster") CuratorCluster hubCluster) {
-        this.hubCluster = hubCluster;
-        this.client = RestClient.createClient(5, 15, true, true);
-    }
-
-    @VisibleForTesting
-    InternalWebhookClient(Client client, CuratorCluster hubCluster) {
+    public InternalWebhookClient(@Named("HubCuratorCluster") CuratorCluster hubCluster,
+                                 @Named("WebhookClient") Client client,
```
lkemmerer: If I'm not mistaken, I think this should be the constant you added up in the named bindings list.

src/test/java/com/flightstats/hub/cluster/CuratorClusterTest.java
```diff
@@ -34,17 +41,19 @@ void testPath() throws Exception {
                 "/SpokeCluster",
                 false,
                 true,
-                new SpokeDecommissionCluster(curator, new SpokeProperties(PropertiesLoader.getInstance())),
+                new SpokeDecommissionCluster(curator, spokeProperties, localHostProperties),
                 new AppProperties(PropertiesLoader.getInstance()),
```
lkemmerer: Would it make sense to use the variables that are newly defined above here for `appProperties` and `spokeProperties`?

src/main/java/com/flightstats/hub/config/properties/LocalHostProperties.java
```diff
@@ -0,0 +1,73 @@
...
+@Slf4j
+public class LocalHostProperties {
+
+    private String uriScheme = "http://";
+    private final int port;
+
+    @Inject
+    public LocalHostProperties(AppProperties appProperties, SystemProperties systemProperties) {
+        this.port =  systemProperties.getHttpBindPort();
+        if (appProperties.isAppEncrypted()) {
```
lkemmerer: In order to avoid having logic in the constructor and a non-final variable above, I think I'd extract this logic into `getUriScheme` and add an `else` that captures the default value.

PR #1200, opened 2019-06-04
Fix /latest query not honoring cache if channel is empty, causing an …

dkleckler:
…S3 hit.

### DIFF
--- a/src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
+++ b/src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
@@ -177,7 +177,7 @@ class ClusterContentServiceTest {
         ContentKey startKey = new ContentKey(TimeUtil.now().minusSeconds(20), "StartKey");
         DirectionQuery query = DirectionQuery.builder().channelName(channelName).stable(true).startKey(startKey).build();
         stubRealChannel();
-        stubCachedEmptyChannelLatest();
+        when(latestContentCache.getLatest(channelName, null)).thenReturn(ContentKey.NONE);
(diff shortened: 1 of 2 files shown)


src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
```diff
@@ -254,6 +283,10 @@ private ContentKey stubCacheLatest(int ageInSeconds, String keyName) {
         return cachedKey;
     }
 
+    private void stubCachedEmptyChannelLatest() {
```
manjula-chikkanagappa: Method is overkill to mock stuff. I would add comments instead

src/test/java/com/flightstats/hub/dao/aws/ClusterContentServiceTest.java
```diff
@@ -17,6 +17,7 @@
 import com.flightstats.hub.util.HubUtils;
 import com.flightstats.hub.util.TimeUtil;
 import lombok.SneakyThrows;
+import org.joda.time.DateTime;
```
manjula-chikkanagappa: Java 8 date library should be favored here. Performance is better with Java 8 dates compared to joda-time.

lkemmerer: I believe we have a task to migrate the hub off joda and onto Java 8 native stuff, but it'll be a pretty big and sensitive change. I'll check to make sure we're tracking that debt.

dkleckler: Yeah, joda's heavily used throughout hub, so leaving this for now for consistency.  Agree it'd be nice to switch over though.

PR #1201, opened 2019-06-05
add content type header to items for replication

Paul-Hess:


lkemmerer:
Test'll get in once I get off my butt and finish up the system test stuff, I'm guessing?

Paul-Hess:
I got this pushed up and ran tests against it, there is a bug, the content-length needs to be reset if the item is compressed. Closing while I fix that.

PR #1203, opened 2019-06-05
Start helmifying system tests

lkemmerer:
This is not ready to merge. I need to make a few tweaks, but having folks take a look at it now should help streamline getting all the system test stuff in. It'll let me incorporate feedback while I continue working and get y'all familiar with the code so that Paul can PR hist tests against this branch.


src/systemTest/java/com/flightstats/hub/kubernetes/PodRestart.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.List;
+
+@Slf4j
+public class PodRestart {
+
+    @SneakyThrows
+    public void execute(String releaseName, List<String> podNames) {
+
+        try (final DefaultKubernetesClient client = new DefaultKubernetesClient()) {
```
chriskessel: A stray ```final```. Actually, there's a lot of them in a lot of the classes in this PR. I'm guessing it's not really new code, but moved from somewhere?

src/systemTest/java/com/flightstats/hub/client/CallbackClientFactory.java
```diff
@@ -0,0 +1,25 @@
+package com.flightstats.hub.client;
+
+import com.google.inject.Inject;
```
manjula-chikkanagappa: We should use javax.inject.Inject for consistency and portability of hub.

src/systemTest/java/com/flightstats/hub/kubernetes/HubLifecycle.java
```diff
@@ -0,0 +1,58 @@
+package com.flightstats.hub.kubernetes;
+
+import com.flightstats.hub.system.config.DependencyInjector;
+import com.flightstats.hub.system.config.PropertyNames;
```
manjula-chikkanagappa: PropertiesName sounds better :) (Each property has one name)

src/systemTest/java/com/flightstats/hub/kubernetes/HubLifecycle.java
```diff
@@ -0,0 +1,58 @@
...
+    private String chartPath;
+
+    @Inject
+    @Named(PropertyNames.HELM_RELEASE_DELETE)
+    private boolean isHelmReleaseDeletable;
+
+    @Inject
+    private ReleaseInstall releaseInstall;
+
+    @Inject
```
manjula-chikkanagappa: 👍

src/systemTest/java/com/flightstats/hub/kubernetes/PodRestart.java
manjula-chikkanagappa: Yes that's me :)

src/systemTest/java/com/flightstats/hub/kubernetes/ReleaseInstall.java
```diff
@@ -0,0 +1,54 @@
...
+            log.info("Hub helm chart location {} ", chartPath);
+            chartBuilder = chartLoader.load(new URL(chartPath));
+        }
+
+        try (final DefaultKubernetesClient client = new DefaultKubernetesClient();
+             final Tiller tiller = new Tiller(client);
+             final ReleaseManager releaseManager = new ReleaseManager(tiller)) {
+
+            final InstallReleaseRequest.Builder requestBuilder = InstallReleaseRequest.newBuilder();
+            requestBuilder.setTimeout(300L);
```
manjula-chikkanagappa: This timeout value needs little more investigation. This is not enough while running from VPN connection.

src/systemTest/java/com/flightstats/hub/model/WebhookCallbackSetting.java
```diff
@@ -0,0 +1,17 @@
+package com.flightstats.hub.model;
+
+import lombok.Builder;
+
+import java.beans.ConstructorProperties;
+
+@Builder
+public class WebhookCallbackSetting {
+    Integer triesUntilSuccess;
```
manjula-chikkanagappa: tryUntilSuccees?

src/systemTest/java/com/flightstats/hub/system/functional/WebhookErrorTest.java
```diff
@@ -0,0 +1,156 @@
...
+import org.junit.jupiter.api.RepeatedTest;
+
+import java.util.Comparator;
+
+import static com.flightstats.hub.util.StringUtils.randomAlphaNumeric;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@Slf4j
+//@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
manjula-chikkanagappa: Did this work? Compiler did not complain that @BeforeAll method is non static?

src/systemTest/java/com/flightstats/hub/system/functional/WebhookErrorTest.java
```diff
@@ -0,0 +1,156 @@
...
+        webhookService.insertAndVerify(webhook);
+    }
+
+    private void verifyHasReceivedErrorForItem(String firstUrl) {
+        callbackService.awaitHubHasCallbackErrorForItemPath(webhookName, firstUrl);
+        assertTrue(callbackService.hasCallbackErrorInHub(webhookName, firstUrl));
+        log.info("Found callback error for first item {} {}", webhookName, firstUrl);
+    }
+
+    @RepeatedTest(3)
```
manjula-chikkanagappa: is this intended?

src/systemTest/java/com/flightstats/hub/kubernetes/PodRestart.java
lkemmerer: I'll go through and clean 'em up.

src/systemTest/java/com/flightstats/hub/kubernetes/ReleaseInstall.java
lkemmerer: I'll probably need to adjust it once I go back to stabilization work. This branch'll probably only install a callback server and not the whole hub ecosystem, and that's much, much faster than all the other stuff.

src/systemTest/java/com/flightstats/hub/model/WebhookCallbackSetting.java
lkemmerer: This matches what's in the callback server. It is the number of tries it'll make, so I think plural makes sense. I'd go for `try` singular if it were a boolean, though.

src/systemTest/java/com/flightstats/hub/system/functional/WebhookErrorTest.java
lkemmerer: The tests pass, so I guess it's unnecessary. I'll delete rather than comment it out.

src/systemTest/java/com/flightstats/hub/system/functional/WebhookErrorTest.java
lkemmerer: Yep! It's the way Junit 5 does the manual loop this file had been doing for its tests.

src/systemTest/java/com/flightstats/hub/system/functional/WebhookErrorTest.java
lkemmerer: I'll double check to make sure the repeated tests are behaving like I expect them to behave.

src/systemTest/java/com/flightstats/hub/system/functional/WebhookErrorTest.java
lkemmerer: Looks like it doesn't behave quite like I expect. I'll see if I can fix this up. Thanks for catching it.

src/systemTest/java/com/flightstats/hub/system/functional/WebhookErrorTest.java
Paul-Hess: RE: static @BeforeAll, from the docs: ```Such methods are inherited (unless they are hidden or overridden) and must be static (unless the "per-class" test instance lifecycle is used).```

lkemmerer: Working on fixing this now. I broke the functionality in other ways, too. :P

src/systemTest/java/com/flightstats/hub/system/functional/WebhookErrorTest.java
```diff
@@ -144,6 +144,13 @@ void testSettingCursorBeyondErrorClearsErrorStateAndContinues() {
         // verify that no errors exist on the hub
         log.info("Verifying that no errors exist on the hub for webhook {}", webhookName);
         assertTrue(callbackService.isErrorListEventuallyCleared(webhookName));
+
+        log.info("deleting " + channelName + " and " + webhookName);
```
manjula-chikkanagappa: This should be parameterized :)
log.info("deleting {} and {}", channelName , webhookName);

src/systemTest/java/com/flightstats/hub/system/functional/WebhookLifecycleTest.java
```diff
@@ -0,0 +1,119 @@
...
+                withStartItem(channelItems.get(4)).
+                withParallelCalls(2);
+        webhookResource.insertAndVerify(webhook);
+        List<String> channelItemsExpected = channelItems.subList(5, channelItems.size());
+        assertTrue(callbackResource.areItemsEventuallySentToWebhook(webhookName, channelItemsExpected));
+    }
+
+    @Test
+    @SneakyThrows
+    void testWebhookWithStartItem_expectItemsInOrder() {
```
Paul-Hess: based on the name of this test I expect it to substantially different from the previous one, as far as I can tell it is only the data itself that changed, and the parallelism, unless there is an additional assertion I think this might work as a parameterized test

lkemmerer: I'll take a look.

build.gradle
```diff
@@ -180,7 +181,7 @@ dependencies {
     systemTestImplementation 'org.junit.jupiter:junit-jupiter:5.4.0'
 
     systemTestCompile('io.netty:netty-tcnative-boringssl-static:2.0.6.Final:' + osdetector.classifier)
-    systemTestCompile('org.microbean:microbean-helm:2.8.2.1.0.5') {
+    systemTestCompile('org.microbean:microbean-helm:2.12.3.0.0.1-SNAPSHOT') {
```
manjula-chikkanagappa: does this snapshot have new feature not available in production version?

lkemmerer: I was having issues getting requirements to work, though if there are concerns I can try downgrading and see if I can sort things out. I was worried that since the library is expecting helm v 2.8 (from early 2018) and kubernetes 1.9 and our helm charts are written against 2.13 and 1.11, it'd be more likely that we'd run into compatibility issues that would be hard to track down.

src/systemTest/java/com/flightstats/hub/system/functional/WebhookLifecycleTest.java
lkemmerer: I'll add an extra order assertion for the test and give it a clearer name. I'll save parameterization for another PR.

build.gradle
```diff
@@ -60,14 +59,18 @@ test {
 }
 
 task systemTest(type: Test, description: 'Runs the system tests.', group: 'Verification') {
+    // Retrieve properties passed in via -P that coincide with system tests. After "systemTest.", names should match those
+    // found in system-test-hub.properties
+    systemProperties project.getProperties().findAll { property ->
```
lkemmerer: Not sure how folks feel about passing in params to gradle tasks like this. I'm happy to use whatever's standard, though. This works, though (see https://ddt-jenkins.pdx.prod.flightstats.io/job/run-hub-system-tests/).

build.gradle
lkemmerer: I was able to downgrade. The issues I had were fixed with changes unrelated to the version of the library I was using.

lkemmerer:
K, AFAIK, this is ready to go. Let me know if there's anything that hasn't been addressed to your satisfaction, @manjula-chikkanagappa and @Paul-Hess.

build.gradle
```diff
@@ -10,7 +10,6 @@ if (project.ext.properties.get('isDevBuild') == null) {
 
 repositories {
     mavenCentral()
-    maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
```
manjula-chikkanagappa: Ah! Thanks for cleaning up this.

build.gradle
manjula-chikkanagappa: what properties are passed here? Can you use system-test.properties file ?

lkemmerer:
I use gradle properties to pass in the hub url for running system tests on Jenkins against whichever cluster is passed in as a parameter. This seemed like a more straightforward method than recreating/modifying the properties file during a test run. The way I implemented it will allow Jenkins to generically override any properties rather than just the hub url, which decreases the number of places we have the property names hard coded (so just the jenkins job, the constants file, and the properties file, rather than all that AND the build.gradle file).

lkemmerer:
It looks like I could use a `-D` (system properties) on the gradle command rather than `-P` (project properties), which I think should take out that block in the build file...

manjula-chikkanagappa:
> It looks like I could use a `-D` (system properties) on the gradle command rather than `-P` (project properties), which I think should take out that block in the build file...

that sounds so much better!

lkemmerer:
I'll double check that it works, but I think it should. Not sure why I didn't do that in the first place.

lkemmerer:
Hm. Switching to `-D` doesn't work without adding `systemProperties System.properties` to the (possibly with the `findAll` unless we want to pass all jvm props over) to the build file. According to this ancient stack overflow answer, it might be because the tests are forked into a new JVM and those properties aren't carried over:  https://stackoverflow.com/questions/21406265/how-to-give-system-property-to-my-test-via-gradle-and-d 
@manjula-chikkanagappa If you prefer using `-D` and `System.properties` instead of project properties, I'm happy to make that change. I don't have a preference.

manjula-chikkanagappa:
> Hm. Switching to `-D` doesn't work without adding `systemProperties System.properties` to the (possibly with the `findAll` unless we want to pass all jvm props over) to the build file. According to this ancient stack overflow answer, it might be because the tests are forked into a new JVM and those properties aren't carried over: https://stackoverflow.com/questions/21406265/how-to-give-system-property-to-my-test-via-gradle-and-d
> @manjula-chikkanagappa If you prefer using `-D` and `System.properties` instead of project properties, I'm happy to make that change. I don't have a preference.

That would be nice Laurie. These are JVM properties and not gradle properties as pointed out in the stackoverflow link.

build.gradle
```diff
@@ -58,10 +58,10 @@ test {
...
 
 task systemTest(type: Test, description: 'Runs the system tests.', group: 'Verification') {
-    // Retrieve properties passed in via -P that coincide with system tests. After "systemTest.", names should match those
+    // Retrieve properties passed in via -D that coincide with system tests. After "systemTest.", names should match those
     // found in system-test-hub.properties
-    systemProperties project.getProperties().findAll { property ->
-        property.getKey().startsWith("systemTest.")
+    systemProperties System.properties.findAll { property ->
+        property.key.startsWith("systemTest.")
     }
```
manjula-chikkanagappa: 👍 This PR looks good to me!

lkemmerer:
got a verbal/slack 👍 from Paul, too.

PR #1204, opened 2019-06-05
Storage tests

Paul-Hess:
will merge this on monday


src/systemTest/resources/system-test-hub.properties
```diff
@@ -3,3 +3,10 @@ helm.chart.path=http://ddt-nexus.flightstats.io/repository/helm/hub.tgz
 #helm.release.delete=false
 hub.url=http://hub-%s.dls.dev.flightstats.io
 callback.url=http://callback-server-%s.dls.dev.flightstats.io
+
+#### s3
+s3.url=http://s3-%s.dls.dev.flightstats.io
+s3.region=us-east-1
+s3.credentials.path=/etc/.aws/credentials
+s3.bucket.name=hub_%s-local
+s3.mock.enabled=false
```
Paul-Hess: localstack is more advanced than other mocks I have used and the pathStyleAccessEnabled flag will only be useful if running against a localhost s3 mock so I am considering this overkill for now and think I should remove it

src/systemTest/java/com/flightstats/hub/system/config/GuiceModule.java
```diff
@@ -16,14 +19,21 @@
...
 
 @Slf4j
 public class GuiceModule extends AbstractModule {
+    private final StringHelper stringHelper = new StringHelper();
+    private final String releaseName = "loopy-joey"; //"ddt-" + stringHelper.randomAlphaNumeric(10).toLowerCase();
     private static final String PROPERTY_FILE_NAME = "system-test-hub.properties";
 
     @Override
     protected void configure() {
+        bind(String.class).annotatedWith(Names.named("test.data")).toInstance("TEST_DATA");
```
Paul-Hess: There is some more cleanup I can do in here, and in a few other files maybe to use the PropertyNames/PropertyLoader strategy for stuff like this

src/systemTest/java/com/flightstats/hub/system/service/ChannelService.java
```diff
@@ -71,6 +85,77 @@ public String addItem(String channelName, Object data) {
...
+        return Arrays.asList(itemPath.split("/"));
+    }
+
+    private Map<DatePathIndex, Integer> getPathDateParts(List<String> pathParts) {
+        Map<DatePathIndex, Integer> dateValues = new HashMap<>();
+        List<Integer> dateParts = pathParts.subList(1, pathParts.size() -1).stream()
+                .filter(StringUtils::isNotBlank)
+                .map(Integer::parseInt)
+                .collect(Collectors.toList());
+        dateValues.put(DatePathIndex.YEAR, dateParts.get(DatePathIndex.YEAR.getIndex()));
```
chriskessel: Might be nice to push the logic into DatePathIndex, so the code would look like:
```
dateValues.put(DatePathIndex.YEAR.extract(dateParts));
dateValues.put(DatePathIndex.MONTH.extract(dateParts));
etc
etc
```
Or even have a method on DatePathIndex that knows how to take a List<Integer> and return a Map:
```
dateValues = DateValueIndex.extractDates(dateParts);
```

src/systemTest/java/com/flightstats/hub/system/service/ChannelService.java
```diff
@@ -71,6 +85,77 @@ public String addItem(String channelName, Object data) {
...
+        return keys;
+    }
+
+    @SneakyThrows
+    private Optional<TimeQuery> getItemByTimeFromLocation(String path, Location location) {
+        List<String> pathParts = getPathParts(path);
+        Map<String, String> keys = getPathKeys(pathParts);
+        Map<DatePathIndex, Integer> dateParts = getPathDateParts(pathParts);
+        Call<TimeQuery> response = channelItemResourceClient.getItemsSecondsPath(keys.get("channelName"),
+                dateParts.get(DatePathIndex.YEAR),
```
chriskessel: It sort of feels like there might be value in an actual DatePath class.

lkemmerer:
It looks like this PR includes some work from master and my branch, maybe? I'm guessing it's because I was stupid and force-pushed while you were basing your work off that. :( Sorry. Let me know if you'd like a hand sorting that out.

src/systemTest/java/com/flightstats/hub/clients/callback/CallbackClientFactory.java
```diff
@@ -7,7 +7,7 @@
 
 public class CallbackClientFactory {
 
-    private final Retrofit retrofitHub;
+    private Retrofit retrofitHub;
```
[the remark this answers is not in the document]
Paul-Hess: find and replace overkill

src/systemTest/java/com/flightstats/hub/system/service/ChannelService.java
Paul-Hess: True

src/systemTest/java/com/flightstats/hub/model/Links.java
```diff
@@ -0,0 +1,12 @@
...
+
+import lombok.Builder;
+import lombok.Value;
+
+@Builder
+@Value
+public class Links {
+    private Link self;
+    private Link channel;
+    private String [] uris;
```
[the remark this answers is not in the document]
Paul-Hess: It feels kind of icky to me to reuse code from the hub in tests that may be testing the validity of that code. I can certainly get over that feeling, is there a convention and a clear line on this?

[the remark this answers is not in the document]
manjula-chikkanagappa: Model classes should be reused in tests. It would be redundant to redefine them in the test and error prone. Data model of the application does not change,.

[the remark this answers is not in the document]
chriskessel: Well, it's more code to maintain, more things to get out of sync if you ever change one of those classes. There's some value in testing with a completely different code base, but that's typically restricted to mission critical things like NASA rover code or some such because it comes at a high cost.

I mean, to go wild on that front, if you really want to avoid using the same code then you'd want to use completely different 3rd party libraries as well. Though, that's not feasible if you're using the same Java core libraries, so then you'd be looking at writing the tests in a completely independent language.
 
Less code is generally better. Less to maintain, less to fix, less bugs created. I'd trust reusing my core domain POJOs like this the same way I'd trust using URI or String.

src/systemTest/java/com/flightstats/hub/model/DatePathIndex.java
```diff
@@ -0,0 +1,21 @@
+package com.flightstats.hub.model;
+
+public enum DatePathIndex {
+    YEAR(0),
```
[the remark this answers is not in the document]
lucasmo: I've never been a huge fan of using `Enum.ordinal()` to express something other than enum sort order. The vast majority of enums have meaningless ordinal values. It's really easy to add another value say, alphabetically or logically, and accidentally change the underlying meaning of some downstream code in a way you didn't intend. 

I suppose a comment could address this as well.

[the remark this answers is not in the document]
chriskessel: True. 

I think if a DatePath class was created to hide all this indexing, then it becomes somewhat moot. The DatePathIndex enum smells like a responsibility that should be hidden inside a class, it's showing us the guts about a defined "schema" for what components exist in a path.

src/systemTest/java/com/flightstats/hub/model/Links.java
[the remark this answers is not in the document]
lkemmerer: I think one of the problems with many of the hub domain objects is that they're not simple POJOs and have strange responsibilities that aren't just modeling the domain. At that point, it feels a little gross letting that production logic leak into system tests, which should be black box. Another problem is that the hub does some super strange things with manual serialization, so some things that you get in a response don't accurate map to a model in the hub. I'd really love to work towards extracting the domain objects into a library of small lombok objects that specify inputs and outputs of the hub and can be used in both the hub and in clients/consumers.

src/systemTest/java/com/flightstats/hub/system/functional/storage/BatchItemStorageTest.java
```diff
@@ -0,0 +1,73 @@
...
+    @AfterEach
+    void cleanup() {
+        channelService.delete(channelName);
+    }
+
+    @Test
+    void batchChannelStorage_itemInSpoke_item() {
+        Awaitility.await()
+                .atMost(Duration.TEN_SECONDS)
+                .until(() -> channelService.getItem(itemUri).equals(testData));
```
lkemmerer: Does this explicitly test that the item's in spoke (i.e. it uses the weird short-term/long-term cache param)? That flag's a little weird and probably only exists for testing. Anyway, if it doesn't check that it's in spoke explicitly, the test name may need a little tweak.

src/systemTest/java/com/flightstats/hub/system/functional/storage/BatchItemStorageTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+import static com.flightstats.hub.util.StringUtils.randomAlphaNumeric;
+import static com.flightstats.hub.system.config.PropertiesName.TEST_DATA;
+
+@Slf4j
+class BatchItemStorageTest extends DependencyInjector {
+    @Inject
```
lkemmerer: Might it make sense to use parameterized tests for BATCH, BOTH, SINGLE tests and have them all in one file? It might make it easier to see the differences in how we expect each scenario to behave (in finding stuff in the hub/spoke and how things are stored in S3). It'd possibly make the tests themselves more obtuse, though, so maybe it's not worth it.

src/systemTest/java/com/flightstats/hub/system/service/ChannelService.java
```diff
@@ -37,15 +47,19 @@ public String getChannelUrl(String channelName) {
...
 
     @SneakyThrows
     public void create(String channelName) {
         log.info("Create channel name {} ", channelName);
+        createCustom(Channel.builder().name(channelName).build());
+    }
 
-        Call<Object> call = channelResourceClient.create(Channel.builder().name(channelName).build());
+    @SneakyThrows
+    public void createCustom(Channel channel) {
```
lkemmerer: This is a trivial suggestion, but what do you thing about renaming `create` to `createWithDefaults` and then making `createCustom` just plain `create`?

src/systemTest/java/com/flightstats/hub/system/service/S3Service.java
```diff
@@ -0,0 +1,132 @@
...
+        S3Object obj = s3Client.getObject(getObjectRequest);
+        try (S3ObjectInputStream content = obj.getObjectContent()) {
+            return handleZip(content);
+        }
+    }
+
+    @SneakyThrows
+    private byte[] getS3Items(String path) {
+        GetObjectRequest getObjectRequest = new GetObjectRequest(bucketName, path);
+        S3Object obj = s3Client.getObject(getObjectRequest);
```
lkemmerer: Does this way ensure the stream/http connection is auto-closed? It looks kinda like the signature that takes a file path to store it /does/ auto-close, but I think you might want to use a try-with-resource starting on this line to ensure that the `Closeable` interface is fully leveraged. Maybe. I can never remember how to use this in the safest way possible.

Paul-Hess: I think you are right, this needs to be try w resources

src/systemTest/java/com/flightstats/hub/system/service/ChannelService.java
Paul-Hess: good point

src/systemTest/java/com/flightstats/hub/system/functional/storage/BatchItemStorageTest.java
Paul-Hess: parameterized this

src/systemTest/java/com/flightstats/hub/system/functional/storage/BatchItemStorageTest.java
Paul-Hess: >> Does this explicitly test that the item's in spoke
It doesn't necessarily, will rename

src/systemTest/java/com/flightstats/hub/model/Links.java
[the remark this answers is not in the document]
Paul-Hess: I've gone through and replaced what I could with the hub core model, Links is not available in that model

src/systemTest/resources/system-test-hub.properties
```diff
@@ -1,17 +1,17 @@
 helm.chart.path=http://ddt-nexus.flightstats.io/repository/helm/hub.tgz
-# helm.release.name=
+helm.release.name=ddt-hessp-hv7y
```
lkemmerer: Probably want to re-comment these properties before you merge.

PR #1210, opened 2019-06-18
suite extension framework

Paul-Hess:


build.gradle
```diff
@@ -63,8 +63,6 @@ task systemTest(type: Test, description: 'Runs the system tests.', group: 'Verif
     systemProperties System.properties.findAll { property ->
         property.key.startsWith("systemTest.")
     }
-    systemProperty 'junit.jupiter.execution.parallel.enabled', 'TRUE'
-    systemProperty 'junit.jupiter.execution.parallel.mode.default', 'CONCURRENT'
```
lkemmerer: I was wondering why tests stopped running in parallel. I put these values in a properties file under system tests instead. Not sure that was the right thing to do.

Paul-Hess: It was, I think before I was getting them extended from the test task setup, wasn't sure why they stopped showing up, and guessed that maybe they were overwritten by setting systemProperties here. It works with the properties file under in this packages resources though and that is more consistent with what we are already doing so I deleted this.

src/systemTest/java/com/flightstats/hub/system/config/GuiceModule.java
```diff
@@ -22,12 +22,15 @@
 
 @Slf4j
 public class GuiceModule extends AbstractModule {
-    private static final String PROPERTY_FILE_NAME = "system-test-hub.properties";
+    private final Properties properties;
+
+    public GuiceModule(Properties properties) {
```
chriskessel: FWIW, you can generate this with lombok by adding ```@AllArgsConstructor``` to the class.

src/systemTest/java/com/flightstats/hub/system/extension/DependencyInjectionExtension.java
```diff
@@ -0,0 +1,23 @@
...
+import com.google.inject.Injector;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import java.util.Optional;
+
+@Slf4j
+public class DependencyInjectionExtension implements BeforeAllCallback {
+    private static final ExtensionContext.Namespace NAMESPACE = ExtensionContext.Namespace.GLOBAL;
```
chriskessel: I'm not sure I see the value in a static final to reference an enum. Why not just reference GLOBAL at the right point in the code?

Paul-Hess: good point, this started as constant because it was to be a placeholder for future custom namespaces, but I'm not sure that is a road we will be able to go down and needs to get cleaned up.

src/systemTest/java/com/flightstats/hub/system/service/CallbackService.java
```diff
@@ -92,6 +93,18 @@ public boolean areItemsEventuallySentToWebhook(String webhookName, List<String>
         }
     }
 
+    public boolean itemsNotSentToWebhook(String webhookName, List<String> channelItems, WebhookType type) {
+        try {
+            int assertAfterWaitSeconds = type.equals(WebhookType.MINUTE) ? 90 : 30;
+            Thread.sleep(assertAfterWaitSeconds * 1000);
```
chriskessel: The variable name already does a good job of making it clear, but I usually try to use the TimeUnit class rather than express the math in the code. E.g:
```
Thread.sleep(TimeUnit.SECONDS(assertAfterWaitSeconds).toMillis());
```

PR #1211, opened 2019-06-18
Pause tests

Paul-Hess:


src/systemTest/java/com/flightstats/hub/system/functional/WebhookPauseTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @AfterEach
+    void cleanup() {
+        channelService.delete(channelName);
+        webhookService.delete(webhookName);
+    }
+
+    @ParameterizedTest
+    @EnumSource(WebhookType.class)
+    void webhookPause_webhookPauseAndUnPause_itemsAreCalledBackAfterUnPause(WebhookType type) {
```
lkemmerer: I'd almost suggest splitting this up into multiple tests (unpaused -> paused and paused -> unpaused) using different channels/webhooks. Do you feel like having it all in one test covers edge cases that wouldn't be caught by splitting up the behavior?

PR #1214, opened 2019-06-26
add more statsd event logging around hub startup

lkemmerer:
We have a statsd event in the shutdown manager (https://app.datadoghq.com/event/stream?tags_execution=and&show_private=true&per_page=30&query=tags%3Acluster%3Aiad.staging%2Crestart%20hub%20priority%3Aall%20status%3Aall&aggregate_up=true&use_date_happened=false&display_timeline=true&from_ts=1561553760000&priority=normal&live=true&is_zoomed=false&status=all&to_ts=1561568160000&is_auto=false&incident=true&only_discussed=false&no_user=false&page=0&bucket_size=240000) , but we don't have anything that tracks startup.

I was debating between "events" and using a counter of some sort so we can create an alert on high rates of restarts. This way follows the existing pattern... What do you think?

### DIFF
--- a/src/main/java/com/flightstats/hub/app/ShutdownManager.java
+++ b/src/main/java/com/flightstats/hub/app/ShutdownManager.java
@@ -46,7 +46,7 @@ public class ShutdownManager {
 
     public boolean shutdown(boolean useLock) throws Exception {
         log.warn("shutting down!");
-        String[] tags = {"restart", "shutdown"};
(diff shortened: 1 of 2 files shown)


Paul-Hess:
event is datadog specific so there may be value in moving away from that pattern and using a counter

lkemmerer:
Cool, I think that'll be more useful. I'll close this and do a new PR that switches over to a pattern we can use with TICK and, more generally, add monitoring/alerting to.

PR #1216, opened 2019-07-01
Metrics around webhook leadership and hub startup/shutdown

lkemmerer:


src/main/java/com/flightstats/hub/app/ShutdownManager.java
```diff
@@ -114,6 +117,11 @@ private void waitForLock() throws Exception {
         }
     }
 
+    private void reportEvent(String eventName) {
+        String[] tags = {"restart", "shutdown"};
+        statsdReporter.incrementCounter(eventName, tags);
+    }
```
manjula-chikkanagappa: This functionality in this method is used in various places and in different forms. would it possible to pull these things in to one place?

lkemmerer: The main reason I extracted it in a couple places was that I didn't want to duplicate the tags in each place I needed to report an event for a particular class. Hiding the tags from the calling method and keeping the tags close to the reporter (vs. a private static final string array) seemed to increase readability. I believe the tags are different in each of the classes that use the pattern, so I didn't see a way to DRY up the behavior any further. What did you have in mind?

manjula-chikkanagappa: An EventReporter which accepts  event name/type and tags. Here the tags will be encapsulated in the object where it belongs.

lkemmerer: Oh! So something like:
```
private final Map<EnumEventName, String[]> eventsToTags = new HashMap(...);
public void reportStartEvent(EnumEventName eventType) {
    statsdReporter.incrementCounter(eventType + ".start", eventsToTag.get(eventType));
}
```
?

PR #1217, opened 2019-07-08
Resiliency test setup

Paul-Hess:
I started this work for max items test, but no longer need for that particular test, instead added a placeholder test to show how it's use until a real test takes it's place soon.


src/systemTest/java/com/flightstats/hub/system/resiliency/PlaceholderTest.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.inject.Inject;
+
+/*
+    A temporary example of how SingletonTestInjectionExtension can be used with custom hub setup
+*/
+@Slf4j
+@ExtendWith(SingletonTestInjectionExtension.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class PlaceholderTest {
+    private static final String CUSTOM_YAML_FOR_HELM = "hub: \n" +
```
Paul-Hess: I considered loading an actual yaml file here, it might make these tests a little cleaner, thoughts?

build.gradle
```diff
@@ -63,6 +63,7 @@ task systemTest(type: Test, description: 'Runs the system tests.', group: 'Verif
     systemProperties System.properties.findAll { property ->
         property.key.startsWith("systemTest.")
     }
+    systemProperty 'kubernetes.websocket.ping.interval', 30000
```
lkemmerer: Is this something you could stick in the properties config file? The only reason the other ones are here are for override purposes...

src/systemTest/java/com/flightstats/hub/kubernetes/ReleaseInstall.java
```diff
@@ -28,33 +28,59 @@ public ReleaseInstall(HelmProperties helmProperties) {
...
+        requestBuilder.setName(getReleaseName());
+        requestBuilder.setWait(true);
+        requestBuilder.setDisableHooks(false);
+        return requestBuilder;
+    }
 
-    @SneakyThrows
-    void install() {
-        log.info("Hub release {} install begins", getReleaseName());
+    private InstallReleaseRequest.Builder configureRequestBuilder() {
```
lkemmerer: can this just call the method below with an empty string as the customYaml?

src/systemTest/java/com/flightstats/hub/system/resiliency/PlaceholderTest.java
manjula-chikkanagappa: It would be nice to maintain these test properties in one place. Setting test properties  manually in multiple places was one of the issues we had in java integration tests resulting in the tight coupling of the tests.

build.gradle
Paul-Hess: I tried that, the properties files don't load the values into System.properties where kubernetes looks for it

lkemmerer: Ah, bummer.

src/systemTest/java/com/flightstats/hub/kubernetes/HelmYamlOverride.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private Map<String, Object> buildClusteredHub() {
+        Map<String, Object> clusteredHub = new HashMap<>();
+        clusteredHub.put(YamlKeys.CLUSTER_HUB_ENABLED, helmProperties.isHubInstallClustered());
+        return clusteredHub;
+    }
+
+    private Map<String, Object> buildHub() {
+        Map<String, Object> hub = new HashMap<>();
```
lkemmerer: Ever since @chriskessel showed off `Map.ofEntries` I've been wishing we were on Java 9. 😞

src/systemTest/java/com/flightstats/hub/system/config/HelmProperties.java
```diff
@@ -1,57 +1,44 @@
...
                           @Named(PropertiesName.HELM_CLUSTERED_HUB) boolean isClusteredHubInstall,
                           @Named(PropertiesName.HELM_RELEASE_DELETE) boolean shouldDeleteRelease,
+                          @Named(PropertiesName.HUB_DOCKER_IMAGE) String hubDockerImage,
                           ServiceProperties serviceProperties) {
         this.releaseName = releaseName;
         this.chartPath = chartPath;
         this.isClusteredHubInstall = isClusteredHubInstall;
         this.shouldDeleteRelease = shouldDeleteRelease;
+        this.hubDockerImage = hubDockerImage;
+        this.hubInstalledByHelm = serviceProperties.getHubUrl().contains(releaseName);
```
lkemmerer: This is trivial, but I try not to do anything except mapping variables to variables in initializers on the off-chance somebody's dumb and passes in a null value somewhere because they know they aren't going to use some property. I may be alone in this opinion and I'm ok with that..

src/systemTest/java/com/flightstats/hub/kubernetes/HelmYamlOverride.java
Paul-Hess: feeling that

Paul-Hess:
> One question that is outside the scope is: what happens to these tests if I have the hub url set to a dev environment or an environment that's already spun up? Do we still create a new install if the configuration for a test isn't default?

It will ignore the install if release name is set to an existing environment just like the helm suite configuration. It assumes the user to set that value to an environment that was built on the expected config.

lkemmerer:
I guess that makes sense. I suppose I'm just worried I'm going to stupidly try to run tests that expect something with a very specialized against iad.dev or something. Which is a PEBKAC issue, not something we should solve with code. ...thanks for letting me think through that here. :P

PR #1220, opened 2019-07-12
Max items system tests

Paul-Hess:


src/main/java/com/flightstats/hub/dao/aws/MaxItemsEnforcerResource.java
```diff
@@ -0,0 +1,40 @@
...
+    @Inject
+    public MaxItemsEnforcerResource(S3Config s3Config) {
+        this.s3Config = s3Config;
+    }
+
+    @PATCH
+    @Path("/{channel}")
+    public Response enforce(@PathParam("channel") String channel) {
+        try {
+            S3Config.S3ConfigLockable s3ConfigLockable = s3Config.new S3ConfigLockable(Collections.emptyList());
```
lkemmerer: I usually try to avoid calling non-static nested classes from outside the class it's defined in...it breaks encapsulation, gets confusing, and tends to indicate that responsibilities are a bit muddled.

src/main/java/com/flightstats/hub/dao/aws/S3Config.java
```diff
@@ -92,10 +93,10 @@ protected Scheduler scheduler() {
 
     }
 
-    private class S3ConfigLockable implements Lockable {
+    class S3ConfigLockable implements Lockable {
```
lkemmerer: Ah, this was an existing thing and you were trying to change as little as possible. I think I'd extract all the max update methods into a top-level class and use it as a dependency in this lockable and its parent. Then this can go back to hiding as private. Does that make sense?

src/systemTest/java/com/flightstats/hub/system/functional/ChannelMaxItemsEnforcementTest.java
```diff
@@ -0,0 +1,106 @@
...
+            batch items are compared with granularity of one minute
+            so if the excess items are of the same minute they will not be deleted
+        */
+        items.add(channelItemCreator.addItem(channelName, TEST_DATA));
+        Thread.sleep(TimeUnit.MINUTES.toMillis(1));
+        items.addAll(channelItemCreator.addItems(channelName, TEST_DATA, 5));
+    }
+
+    private void waitUntilAllItemsAreInS3(ChannelType type) {
+        Awaitility.await().pollInterval(Duration.FIVE_SECONDS)
```
lkemmerer: Might it make sense to use the LOCATION query param against the hub api instead of reaching into s3 for this? I feel like we might be able to avoid breaking long-term storage encapsulation without losing anything here.

src/systemTest/java/com/flightstats/hub/system/functional/ChannelMaxItemsEnforcementTest.java
```diff
@@ -0,0 +1,106 @@
...
+                .atMost(new Duration(3, TimeUnit.MINUTES))
+                .until(() -> s3Service.confirmItemsInS3(type, items, channelName) &&
+                        items.stream().allMatch(path -> channelItemRetriever.getItem(path).isPresent()));
+    }
+
+    private String setupTest(ChannelType type) {
```
lkemmerer: `createChannelAndAddTooManyItems` might be a better name for this method. I think that might indicate this method is hiding too much stuff though. I'd break it into something like a `createChannelWithMaxItems(5)`, `addItem` (which saves the path), `wait`, `addItems(5)` step, especially since there's only one test method in this class.

src/systemTest/java/com/flightstats/hub/clients/hub/channel/ChannelItemResourceClient.java
```diff
@@ -43,6 +43,14 @@
                                               @Path("s") int second,
                                               @Query("location") Location location);
 
+    @GET("/channel/{channelName}/{Y}/{M}/{D}/{h}")
+    Call<TimeQueryResult> getItemsHourPath(@Path("channelName") String channelName,
+                                              @Path("Y") int year,
```
chriskessel: Nit, though I guess it agreement with the other time path endpoints. Why abbreviate Y, M, D, etc here, but use full names on other end points (like directionalItems). And then, why lowercase "h", but upper case the others? I thought maybe it was because it was the abbreviations used by DateTimeFormatter, but that doesn't match up either.

Like I said, it's a nit, maybe not worth noting, just struck me as odd/inconsistent.

src/systemTest/java/com/flightstats/hub/system/functional/ChannelMaxItemsEnforcementTest.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @SneakyThrows
+    private void addItemsOverTime() {
+        /*
+            batch items are compared with granularity of one minute
+            so if the excess items are of the same minute they will not be deleted
+        */
+        items.add(channelItemCreator.addItem(channelName, TEST_DATA));
+        Thread.sleep(TimeUnit.MINUTES.toMillis(1));
```
chriskessel: Bummer to wait a minute on a test doing nothing :(. Not sure what a good solution is though. Maybe an optional channelItemCreator constructor arg or backdoor setter to hack how it decides the time of the item added, so you do something like:
```
items.add(channelItemCreator.addItem(channelName, TEST_DATA))
channelItemCreator.hackAddThisTimeToAddItem(TimeUnits.MINUTES.toMillis(1));
items.addAll(....);
```

src/systemTest/java/com/flightstats/hub/clients/hub/channel/ChannelItemResourceClient.java
lkemmerer: The directional items thing is my fault. I got tired of parsing out an existing content key back into args just to construct 'em back into the same format content key as it started out.

src/systemTest/java/com/flightstats/hub/system/functional/ChannelMaxItemsEnforcementTest.java
lkemmerer: I was going to suggest using a historical channel, but I think that prevents us from changing over to have a max items setting...which means we'd have to mess with dynamo directly to update the config.

lucasmo: Without knowing much about the hub's codebase, is there a way to use a Clock object? ie, in production you use the system clock, but in test you use a mutable clock?

zradick: +1 to Lucas' suggestion.
We started from the Guava Ticker because it fit some other use cases but a convertible time source that is used globally across the application has a lot of benefits.  Happy to share pointers to our various implementations if desired but if I was doing it today from scratch I would probably expose getInstant() and push all use towards that contract myself.

lucasmo: By Clock object I mean a `java.time.Clock`

lkemmerer: I think moving that way would be great, but I'm hesitant to mess with the way hub deals with time before we have holistic tests that make sure we don't break the current implementation or emergent(/undocumented) behavior.

Paul-Hess: > I think moving that way would be great, but I'm hesitant to mess with the way hub deals with time before we have holistic tests that make sure we don't break the current implementation or emergent(/undocumented) behavior.

I agree with @lkemmerer about getting into tinkering with the hub's time implementation, though long term those ideas would make testing things like this more ergonomic and stable. We have had a conversation offline about this as well. In this particular case I think doing too much to work around this would amount to distracting from what is a (albeit minor) bug.

src/main/java/com/flightstats/hub/dao/aws/S3Config.java
Paul-Hess: makes sense, shied away from changing too much in the first iteration but changing this was easy without breaking anything

src/systemTest/java/com/flightstats/hub/system/functional/ChannelMaxItemsEnforcementTest.java
Paul-Hess: I went back and forth on this I think that would be fine

src/main/java/com/flightstats/hub/dao/aws/MaxItemsEnforcer.java
```diff
@@ -0,0 +1,89 @@
...
+                            ChannelService channelService,
+                            @Named("ChannelConfig") Dao<ChannelConfig> channelConfigDao) {
+        this.contentRetriever = contentRetriever;
+        this.channelService = channelService;
+        this.channelConfigDao = channelConfigDao;
+    }
+
+    void updateMaxItems(Iterable<ChannelConfig> configurations) {
+        log.info("iterating channels for max items update");
+        for (ChannelConfig config : configurations) {
```
lkemmerer: what do you think about
```
configurations.stream()
    .filter(config -> config.getMaxItems() > 0)
    .filter(config -> !config.getKeepForever())
    .ifPresent(this::updateMaxItems);
```
?
That looks suspiciously like the Optional filter below, but that's probably A-OK.

src/main/java/com/flightstats/hub/dao/aws/MaxItemsEnforcer.java
```diff
@@ -0,0 +1,89 @@
...
+            if (config.getMaxItems() > 0 && !config.getKeepForever()) {
+                updateMaxItems(config);
+            }
+        }
+    }
+
+    @SneakyThrows
+    void updateMaxItems(String channelName) {
+        Optional<ChannelConfig> channelConfig = Optional.ofNullable(channelConfigDao.get(channelName));
```
lkemmerer: As a stylistic preference, I'd probably chain this with the next line instead of using an intermediate variable:
```Optional.ofNullable(...)
    .filter(....)....
```

src/main/java/com/flightstats/hub/dao/aws/MaxItemsEnforcer.java
```diff
@@ -0,0 +1,89 @@
...
+        channelConfig.filter(config -> config.getMaxItems() > 0 && !config.getKeepForever())
+                .ifPresent(this::updateMaxItems);
+    }
+
+    private void updateMaxItems(ChannelConfig config) {
+        String name = config.getDisplayName();
+        log.info("updating max items for channel {}", name);
+        ActiveTraces.start("S3Config.updateMaxItems", name);
+        Optional<ContentKey> optional = contentRetriever.getLatest(name, false);
```
lkemmerer: I'm slowly realizing that I'm probably making suggestions for code you just moved and didn't write. You don't have to make these changes, but I'm still going to make the comments.

This'd probably be another good candidate to chain optional/filter/ifPresent

src/systemTest/java/com/flightstats/hub/system/functional/ChannelMaxItemsEnforcementTest.java
```diff
@@ -38,60 +39,87 @@
...
-                .until(() -> s3Service.confirmItemNotInS3(type, path, channelName));
+                .until(() -> {
+                    if (type.equals(ChannelType.BOTH)) {
+                        return !confirmInLocation(path, Location.LONG_TERM_SINGLE) &&
+                                !confirmInLocation(path, Location.LONG_TERM_BATCH);
+                    } else {
+                        Location location = type.equals(ChannelType.BATCH) ? Location.LONG_TERM_BATCH : Location.LONG_TERM_SINGLE;
+                        return !confirmInLocation(path, location);
+                    }
+                });
```
lkemmerer: Is this asserting stuff? I think you can add a second param to `.until` that takes a hamcrest matcher and then it would be.

src/systemTest/resources/system-test-hub.properties
```diff
@@ -1,7 +1,7 @@
 helm.chart.path=http://ddt-nexus.flightstats.io/repository/helm/hub.tgz
-# helm.release.name=
```
lkemmerer: you'll probably want to revert this file before merging.

PR #1222, opened 2019-07-17
fixes s3 stream not adding connection back to pool

Paul-Hess:
PDXDDT-838

### DIFF
--- a/src/main/java/com/flightstats/hub/dao/aws/S3AccessMonitor.java
+++ b/src/main/java/com/flightstats/hub/dao/aws/S3AccessMonitor.java
@@ -89,6 +89,7 @@ public class S3AccessMonitor {
             return CompletableFuture.supplyAsync(() -> {
                 try (S3Object s3Object = hubS3Client
                         .getObject(new GetObjectRequest(bucketName, key()))) {
+                    s3Object.getObjectContent().read(new byte[s3Object.getObjectContent().available()]);
                     return s3Object.getObjectMetadata().getVersionId();
                 } catch (IOException e) {
                     log.info("error closing connection to s3", e);
--- a/src/main/java/com/flightstats/hub/dao/aws/S3BatchContentDao.java
+++ b/src/main/java/com/flightstats/hub/dao/aws/S3BatchContentDao.java
@@ -29,7 +29,6 @@ import com.flightstats.hub.model.TimeQuery;
 import com.flightstats.hub.util.TimeUtil;
 import com.google.common.io.ByteStreams;
 import com.google.inject.Singleton;
-import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.joda.time.DateTime;
@@ -153,7 +152,6 @@ public class S3BatchContentDao implements ContentDao {
         return builder.build();
     }
 
-    @SneakyThrows
     private ZipInputStream getZipInputStream(String channel, ContentPathKeys minutePath) {
         ActiveTraces.getLocal().add("S3BatchContentDao.getZipInputStream");
         long start = System.currentTimeMillis();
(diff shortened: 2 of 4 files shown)


lkemmerer:
I think I was seeing this warning when the hub received timeouts while it was processing stuff on the bulk endpoint, too....somewhere in `ZipInputStream`, I think, we're not using a try-with-resources where we might need to. That's probably a different ticket, though...

src/main/java/com/flightstats/hub/dao/aws/S3AccessMonitor.java
```diff
@@ -89,6 +89,7 @@ private String key() {
             return CompletableFuture.supplyAsync(() -> {
                 try (S3Object s3Object = hubS3Client
                         .getObject(new GetObjectRequest(bucketName, key()))) {
+                    s3Object.getObjectContent().read(new byte[s3Object.getObjectContent().available()]);
```
lkemmerer: would calling `s3Object.getObjectContent().close()` do roughly the same thing?

Paul-Hess: that would throw, if we explicitly close the connection without reading all the bytes out it blows up.

Paul-Hess:
> I think I was seeing this warning when the hub received timeouts while it was processing stuff on the bulk endpoint, too....somewhere in `ZipInputStream`, I think, we're not using a try-with-resources where we might need to. That's probably a different ticket, though...

I remembered that I had come across those as I was searching for the culprit. I added a couple simple fixes for those to this as well

src/main/java/com/flightstats/hub/dao/aws/S3AccessMonitor.java
lucasmo: This makes me leery given the unknown size of the bucket. Is the goal here to read the entire stream and discard it, much like an http request? In this case, available() is what's currently in the buffer, not the actual length of the document.

I'd do something like have a 4k buffer and read into it repeatedly until EOF.

src/main/java/com/flightstats/hub/dao/aws/S3BatchContentDao.java
```diff
@@ -152,13 +153,15 @@ private Content getContent(ContentKey key, ZipInputStream zipStream, ZipEntry ne
...
+    @SneakyThrows
     private ZipInputStream getZipInputStream(String channel, ContentPathKeys minutePath) {
         ActiveTraces.getLocal().add("S3BatchContentDao.getZipInputStream");
         long start = System.currentTimeMillis();
         try {
             GetObjectRequest request = new GetObjectRequest(bucketName, getS3BatchItemsKey(channel, minutePath));
-            S3Object object = s3Client.getObject(request);
-            return new ZipInputStream(new BufferedInputStream(object.getObjectContent()));
+            try (S3Object object = s3Client.getObject(request)) {
+                return new ZipInputStream(new BufferedInputStream(object.getObjectContent()));
```
lucasmo: This is going to close the stream on exit.

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/S3Object.java#L223

src/main/java/com/flightstats/hub/dao/aws/S3AccessMonitor.java
lucasmo: (If you have IOUtils available, you can readFully into a NullOutputStream to save on code, I believe)

src/main/java/com/flightstats/hub/dao/aws/S3LargeContentDao.java
```diff
@@ -199,18 +199,20 @@ private Content getS3Object(String channelName, ContentKey key) throws IOExcepti
...
+            try (S3Object object = s3Client.getObject(request)) {
+                ObjectMetadata metadata = object.getObjectMetadata();
+                Map<String, String> userData = metadata.getUserMetadata();
+                Content.Builder builder = Content.builder();
+                String type = userData.get("type");
+                if (!type.equals("none")) {
+                    builder.withContentType(type);
+                }
+                builder.withContentKey(key);
+                builder.withStream(object.getObjectContent());
```
lucasmo: I think this is going to suffer from the same problem of closing the stream in the returned object.

src/main/java/com/flightstats/hub/dao/aws/S3BatchContentDao.java
Paul-Hess: True, tests are broken.

PR #1223, opened 2019-07-25
Tests for Zookeeper's last channel item state as well as the way it affects the `/latest` endpoint

lkemmerer:
This is a little weird and hacky, so I'm open to a nicer way to test this stuff. 

Tried to break up the PR so it's structured with all the dependencies for the new tests first, then the actual tests. (It might be helpful to have the last commit open in a separate window so you can go back and forth to see where stuff is used.)


src/systemTest/java/com/flightstats/hub/model/ChannelConfigExpirationSettings.java
```diff
@@ -0,0 +1,73 @@
...
+    boolean keepForever;
+    long ttlDays;
+    DateTime mutableTime;
+    long maxItems;
+
+    public Optional<DateTime> getMutableTime() {
+        return Optional.ofNullable(mutableTime);
+    }
+
+    public static class ChannelConfigExpirationSettingsBuilder {
```
Paul-Hess: What do you think of lombok-ing this and moving the validation to another class that is the one that is actually consumed, one that essentially takes the arguments, creates an instance with a builder, then validates that. Hmm, then I guess that would have to extend this... NM

Paul-Hess: Is there a custom validation @annotation out there? that would be nice. Found this https://dzone.com/articles/create-your-own-constraint-with-bean-validation-20

src/systemTest/java/com/flightstats/hub/system/functional/ChannelLatestUpdatedStateTest.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @AfterEach
+    void cleanup() {
+        channelConfigService.delete(channelName);
+    }
+
+    @Test
+    void channelWithNoItem_hasStateValueOfEpochNone_and_hasNoChannelLatest() {
+        channelConfigService.createWithDefaults(channelName);
```
Paul-Hess: I think this and line 60 can be combined into the before each setup

src/systemTest/java/com/flightstats/hub/system/functional/ChannelLatestUpdatedStateTest.java
```diff
@@ -0,0 +1,161 @@
...
+        ChannelItemWithBody mostRecentItem = channelItems.last();
+
+        ChannelConfig updatedConfig = ChannelConfig.builder()
+                .name(channelName)
+                .ttlDays(5)
+                .build();
+        channelConfigService.updateExpirationSettings(updatedConfig);
+
+        assertChannelLatestUrlEquals(mostRecentItem, true);
+        // TODO: this never gets set because it was inserted as an historical item. We should decrease spoke TTL and wait 'til it passes. like a chump.
```
Paul-Hess: TIME_MACHINE_MODE

Paul-Hess: I think we should put a SPIKE ticket in to look into the time/impact of setting up a hub mode to run with mutable time settings

lkemmerer: whoops, I forgot to take this todo out. I hadn't rebased onto your fancy new custom release stuff where i could change the ttl...though i think that would fail on CI, since ttl's not going to change on dev.

lkemmerer: I like the idea of adding a spike for that. I briefly considered the idea of setting up different versions of the hub with stuff already in spoke and/or s3 when the release is installed, but that seemed overly complicated...and again, wouldn't work on our normal dev clusters.

src/systemTest/java/com/flightstats/hub/model/ChannelConfigExpirationSettings.java
lkemmerer: This seemed to be the way that the lombok docs suggested validation be done. I'll see if I can find the particular page I was using as a guide... and I'll take a look at your link.

lkemmerer: I lied about the example being from Lombok maintainers. I might be able to override the actual `build()` function to do the validation there...? The constraint annotation route seems a little heavy-handed for this tiny use case, but I can likely be convinced otherwise.

src/systemTest/java/com/flightstats/hub/system/functional/ChannelLatestUpdatedStateTest.java
lkemmerer: I had it that way originally, but some tests use an historical channel and some don't and it was annoying to have to create a second, different channel with a different channel name. this way it's a bit more explicit.

src/systemTest/java/com/flightstats/hub/model/ChannelConfigExpirationSettings.java
```diff
@@ -0,0 +1,73 @@
...
+        private boolean keepForever;
+        private long ttlDays;
+        private DateTime mutableTime;
+        private long maxItems;
+
+        public ChannelConfigExpirationSettingsBuilder keepForever(boolean keepForever) {
+            this.keepForever = keepForever;
+            validate();
```
chriskessel: I'd think the validate would be at the ```.build()``` time rather than each setter. That also allows it to do validation for combinations (e.g. a && b must be set) that wouldn't be possible at individual setter time.

Also, less code, you'd only override build() rather than every setter.

src/systemTest/java/com/flightstats/hub/model/ZookeeperNode.java
```diff
@@ -0,0 +1,14 @@
...
+import lombok.Value;
+
+import java.util.List;
+
+@Builder
+@Value
+public class ZookeeperNode {
+    Links _links;
+    ZookeeperNodeData data;
+    List<ZookeeperChildNode> children;
```
chriskessel: In a effort to avoid null collections and try to ensure they're always just empty length collections, I've started doing this:
```
Links _links;
ZookeeperNodeData data;
@Builder.Default
List<ZookeeperChildNode> children = new ArrayList<>();
```
Figured I'd mention it in case you'd find it useful. I suppose it's generally useful, you could default a boolean to true for instance, but I generally find I want it for collections.

src/systemTest/java/com/flightstats/hub/model/ChannelConfigExpirationSettings.java
lkemmerer: I was worried `build()` wouldn't work right if I defined it here, but that worry is probably silly. I'll move it.

src/systemTest/java/com/flightstats/hub/model/ZookeeperNode.java
lkemmerer: Ooh, good idea. I remember having done some hacky things to avoid null collections in the past. This is nice and simple.

src/systemTest/java/com/flightstats/hub/model/ChannelConfigExpirationSettings.java
```diff
@@ -21,53 +21,36 @@
...
-        }
+    public static Builder builder() {
+        return new Builder();
+    }
 
-        public ChannelConfigExpirationSettingsBuilder maxItems(long maxItems) {
-            this.maxItems = maxItems;
-            validate();
-            return this;
+    public static class Builder extends ChannelConfigExpirationSettingsBuilder {
```
chriskessel: You shouldn't have to do an extends. You should just be able to do this:
```
    public static class ChannelConfigExpirationSettingsBuilder {
        public ChannelConfigExpirationSettings build() {
            validate(); // Could have a validate() on the builder or on the class. I'd put it on the builder.
            return new ChannelConfigExpirationSettings(foo, bar, baz);
        }
  
        private void validate() {
            ...validate foo, bar, and baz
        }
    }
```
If I need to replace the Builder generated code, I'll usually use IntelliJ's Refactor->Delombok to generate the code, then strip out everything except what I want to replace. Lombok is (usually) smart enough to generate everything it doesn't see you've already implemented.

lkemmerer: When I don't do an extends, but I keep @Builder on there, it complains that `build()` is already defined. At that point if I want to avoid an extends, I have to delombokify what seems like most of the builder behavior. :-/

lkemmerer: I'll mess with it a bit to see if I'm missing something.

lkemmerer: ...it's a test class and I think I'm done fighting with lombok for today. I'll make another go at it if I touch it again.

chriskessel: Here's an example that I think does what you're looking for:
```
@Value
@NoArgsConstructor(force = true)
@AllArgsConstructor
@Builder
public class Test {
    String a;
    String b;

    public static void main(String[] args) {
        Test t = Test.builder().b("bar").build();
        System.out.println(t);
    }
    static class TestBuilder {
        public Test build() {
            validate();
            return new Test(this.a, this.b);
        }

        private void validate() {
            if (a == null) throw new RuntimeException("Hey, you have to set a!");
        }
    }
}
```

PR #1227, opened 2019-08-15
Default batch

Paul-Hess:


src/main/java/com/flightstats/hub/model/ChannelConfig.java
```diff
@@ -93,15 +93,21 @@ private ChannelConfig(String name,
             this.ttlDays = ttlDays;
             this.maxItems = maxItems;
         }
-
         if (isBlank(storage)) {
-            this.storage = SINGLE.name();
```
lkemmerer: While you're in here, can you extract the storage type choice logic into a method that returns the type? Then you'd only have to do `this.storage = getStorageType()` or something similar, which sets storage just once. That'd move the historical tagging back out to where it was, but I think right now it gets lost in all the storage logic.

src/test/java/com/flightstats/hub/dao/aws/S3ConfigStrategyTest.java
```diff
@@ -39,7 +39,8 @@ private void verify(int channels, int iterations) {
         for (int i = 0; i < channels; i++) {
             String name = StringUtils.randomAlphaNumeric(10);
             channelConfigs.add(ChannelConfig.builder().name(name).build());
-            names.add(name);
+            String batchedName = name + "Batch";
```
lkemmerer: Do we have a centralized location of this magic-ish string format? I'm guessing we use it in more than one place...

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
```diff
@@ -169,14 +171,47 @@ void testEnforceChannelRetention() {
...
+                .storage(ChannelType.SINGLE.name())
+                .name("defaults").build();
+
+        DateTime mutableTime = TimeUtil.now();
+        ChannelConfig channelConfig = defaults.toBuilder().mutableTime(mutableTime).storage(ChannelType.SINGLE.name()).build();
+        assertEquals(mutableTime, channelConfig.getMutableTime());
+    }
+
+    @Test
+    void testMutableTimeFailOnUpdateWithoutUpdateStorage() {
```
lkemmerer: While this is a reasonable test, I think it's more likely that we'd start with a SINGLE mutable time channel and then try to change storage type to BATCH. I believe we already prevent changing between a historical channel and something with maxItems or a TTL... Maybe add a test for the first case as well?

src/test/java/com/flightstats/hub/webhook/WebhookRetryerTest.java
```diff
@@ -16,6 +18,7 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @ExtendWith(MockitoExtension.class)
+@Execution(ExecutionMode.SAME_THREAD)
```
lkemmerer: I think you'd mentioned why you had to add this mode, but I don't remember the failure/error you encountered. Was it a mockito issue?

src/main/java/com/flightstats/hub/model/ChannelConfig.java
```diff
@@ -93,15 +93,21 @@ private ChannelConfig(String name,
...
+            this.storage = BATCH.name();
         } else {
             this.storage = StringUtils.upperCase(storage);
```
lucasmo: What happens if we put some garbage in `storage`? Is that validated elsewhere?

src/main/java/com/flightstats/hub/model/ChannelConfig.java
```diff
@@ -93,15 +93,21 @@ private ChannelConfig(String name,
...
         }
 
+        if (isHistorical()) {
+            tags.add(HISTORICAL.toString());
+            if (StringUtils.isNotBlank(storage) && !storage.equals(SINGLE.name())) {
```
lucasmo: Can this test the value of this.storage so it's guaranteed to be in sync?

src/main/java/com/flightstats/hub/model/ChannelConfig.java
```diff
@@ -93,15 +93,21 @@ private ChannelConfig(String name,
...
+                throw new InvalidRequestException(storage + " storage is not supported by the historical channel/mutable time api");
```
lucasmo: It feels weird to have a constructor throw an `InvalidRequestException`. Is this class only used when creating new channels, or is it tied to deserializing state that's stored in dynamo? Is it possible this could blow up if there was some weird config stored in dynamo that shouldn't have been possible but the check didn't exist before?

src/main/java/com/flightstats/hub/model/ChannelConfig.java
lkemmerer: I think we do have validation elsewhere. I'm wondering if that exception that we throw could be moved there. I'll try to figure out where we do that. One of the services, maybe?

lkemmerer: There's a `ChannelValidator` class that might be the right place to put that sort of thing... and it does check for a valid storage type.

src/main/java/com/flightstats/hub/model/ChannelConfig.java
Paul-Hess: This check already happened, it happened at a different layer, in ChannelValidator which runs after this builder is built. So it ~~shouldn't~~ have been possible to have that state pre-existing in dynamo, I am working on an alternative strategy to throwing this right here because it is very gross

src/test/java/com/flightstats/hub/webhook/WebhookRetryerTest.java
Paul-Hess: I think I found the problem, another unrelated but heavily mocked test had an unnecessary stub that caused mockito to throw (though it only threw the error sometimes, depending on random test order) removing that stub fixed these failing indeterminately in concurrent mode

src/test/java/com/flightstats/hub/model/ChannelConfigTest.java
Paul-Hess: this and the test case you suggested have moved to the channel validator test where the error now originates

src/test/java/com/flightstats/hub/dao/aws/S3ConfigStrategyTest.java
Paul-Hess: We didn't really, so I added it

PR #1228, opened 2019-08-19
remove influxdb reporter from services, make datadog client primary

Paul-Hess:


src/main/java/com/flightstats/hub/config/ClusterServicesRegistration.java
```diff
@@ -75,7 +72,9 @@ private void registerServices(List<Service> services, HubServices.TYPE type) {
     }
 
     private List<Service> getBeforeHealthCheckServices() {
-        List<Service> services = new ArrayList<>(Arrays.asList(influxdbReporterLifecycle, statsDReporterLifecycle));
+        List<Service> services = new ArrayList<>();
```
lkemmerer: Could you use a `Collections.singletonList` here?

Paul-Hess: singletonList is scoped to having enough memory for one item it's entire lifecycle correct? It does not show in the diff here but more items are conditionally added to the List later

lkemmerer: Oh, duh. I didn't look at the rest of the code.

PR #1233, opened 2019-10-11
Add prefix to s3 lifecycles rules crated by hub

manjula-chikkanagappa:
Add prefix to s3 bucket lifecycle rules created by the HUB. This will help us to distinguish the s3 bucket lifecycle rules created externally (Outside HUB. From terraform code.)

### DIFF
--- a/src/main/java/com/flightstats/hub/dao/aws/S3ConfigStrategy.java
+++ b/src/main/java/com/flightstats/hub/dao/aws/S3ConfigStrategy.java
@@ -12,12 +12,14 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 @Slf4j
 class S3ConfigStrategy {
 
-    public final static String BATCH_POSTFIX = "Batch";
-    public final static String SINGLE_POSTFIX = "";
+    final static String BATCH_POSTFIX = "Batch";
+    final static String SINGLE_POSTFIX = "";
+    final static String BUCKET_LIFECYCLE_RULE_PREFIX = "HUB_";
 
     static List<BucketLifecycleConfiguration.Rule> apportion(Iterable<ChannelConfig> channelConfigs, DateTime timeForSharding, int max) {
         List<BucketLifecycleConfiguration.Rule> rules = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


src/main/java/com/flightstats/hub/dao/aws/S3ConfigStrategy.java
```diff
@@ -82,8 +84,26 @@ static String getChannelTypedName(ChannelConfig config, String postfix) {
...
+     * The lifecycle rules set in terraform code cleans up all the delete markers and previous versions of the object in the S3 bucket.
+     **/
+    static List<BucketLifecycleConfiguration.Rule> getNonHubBucketLifecycleRules(BucketLifecycleConfiguration config) {
+        List<BucketLifecycleConfiguration.Rule> currentBucketLifecycleRules = config.getRules();
+        List<BucketLifecycleConfiguration.Rule> filteredRules =
+                currentBucketLifecycleRules
+                        .parallelStream()
+                        .filter(rule -> !rule.getId().contains(BUCKET_LIFECYCLE_RULE_PREFIX))
+                        .collect(Collectors.toList());
+        return filteredRules.size() < 10 ? filteredRules : filteredRules.subList(0, 10);
```
chriskessel: If you're going to chop it down to 10 anyway, you can use limit(10) on the stream.

src/main/java/com/flightstats/hub/dao/aws/S3ConfigStrategy.java
```diff
@@ -82,8 +84,26 @@ static String getChannelTypedName(ChannelConfig config, String postfix) {
...
+     * The lifecycle rules created in hub only expires the objects based on TTL days configuration in hub channel.
+     * The expiration action in S3 adds delete marker to current object and does not delete it.
+     * The lifecycle rules set in terraform code cleans up all the delete markers and previous versions of the object in the S3 bucket.
+     **/
+    static List<BucketLifecycleConfiguration.Rule> getNonHubBucketLifecycleRules(BucketLifecycleConfiguration config) {
+        List<BucketLifecycleConfiguration.Rule> currentBucketLifecycleRules = config.getRules();
+        List<BucketLifecycleConfiguration.Rule> filteredRules =
+                currentBucketLifecycleRules
+                        .parallelStream()
+                        .filter(rule -> !rule.getId().contains(BUCKET_LIFECYCLE_RULE_PREFIX))
```
lucasmo: I would use `.startsWith()` (I think?) instead of `.contains()` to not get false positives on rules that have "HUB_" somewhere inside them - if that's possible.

manjula-chikkanagappa: that's logical.

src/main/java/com/flightstats/hub/dao/aws/S3ConfigStrategy.java
```diff
@@ -82,8 +84,25 @@ static String getChannelTypedName(ChannelConfig config, String postfix) {
     private static BucketLifecycleConfiguration.Rule createRule(ChannelConfig config, String id) {
         return new BucketLifecycleConfiguration.Rule()
                 .withFilter(new LifecycleFilter(new LifecyclePrefixPredicate(id + "/")))
-                .withId(id)
+                .withId(BUCKET_LIFECYCLE_RULE_PREFIX.concat(id))
```
lkemmerer: It might make sense to move this concatenation up to `getChannelTypedName` where we're already doing some string manipulation on the `id`. That might merit a rename of that method.

manjula-chikkanagappa: I thought about this. There could many callers of this method createRule(). Adding prefix logic here  ensures the prefix is added every time a lifecycle rule is created.

manjula-chikkanagappa:
> Looks good to me. It's unfortunate that the tests around this behavior are obtuse ...and that the code's not structured in a way that'd make your change easy to test, but getting it testable will require a bit of work.
> How will the migration from non-prefixed configuration rules to prefixed ones work? It looks like this'd eventually roll everything over to the new naming convention, but we wouldn't be able to add any real non-hub rules until all the old stuff is phased out.

Yes. I have meeting set up with joe and rich to discuss migration strategies.

PR #1234, opened 2020-01-17
Reduce number of unique tags we send to DataDog.

lkemmerer:
One of our dev clusters had a spike in s3 errors and Datadog contacted us to say that we added 50k unique tags to our metrics overnight. This PR removes unique request IDs and s3 item keys from the tags we send with our metrics. Instead, I log them as part of an error message.

### DIFF
--- a/src/main/java/com/flightstats/hub/dao/aws/HubS3Client.java
+++ b/src/main/java/com/flightstats/hub/dao/aws/HubS3Client.java
@@ -1,6 +1,7 @@
 package com.flightstats.hub.dao.aws;
 
 import com.amazonaws.AmazonWebServiceRequest;
+import com.amazonaws.ResponseMetadata;
 import com.amazonaws.SdkClientException;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.S3ResponseMetadata;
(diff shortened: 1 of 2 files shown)


src/main/java/com/flightstats/hub/dao/aws/HubS3Client.java
```diff
@@ -66,7 +68,7 @@ InitiateMultipartUploadResult initiateMultipartUpload(InitiateMultipartUploadReq
         try {
             return s3Client.initiateMultipartUpload(request);
         } catch (SdkClientException e) {
-            countError(e, request, "initiateMultipartUpload", Arrays.asList("bucket:" + request.getBucketName(), "key:" + request.getKey()));
+            countError(e, request, "initiateMultipartUpload", Collections.singletonList(request.getKey()));
```
chriskessel: FWIW, with JDK 9+ you can use `List.of(...)` and it's smart about optimizing for the number of passed elements.

lkemmerer: I keep forgetting about that and letting IntelliJ convert from `asList`.

lkemmerer: And also forgetting that that hub uses Java 8. :(

PR #1235, opened 2020-01-23
Fix webhook config dynamo scans

lkemmerer:
This adds the fix to the bug that Lucas discovered around webhook config retrieval. The bug was that once the returned list of records reached a certain size and required paging, the method would enter an infinite loop.

Since I was adding tests (based on channel config dao tests), I pulled over a fix that Damon made to channel configs, where improperly constructed dynamo records are dropped rather than throwing an error and causing the hub to behave very badly.

### DIFF
--- a/src/main/java/com/flightstats/hub/dao/aws/DynamoWebhookDao.java
+++ b/src/main/java/com/flightstats/hub/dao/aws/DynamoWebhookDao.java
@@ -21,7 +21,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.stream.Stream;
 
 @Slf4j
 public class DynamoWebhookDao implements Dao<Webhook> {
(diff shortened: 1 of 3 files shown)


src/main/java/com/flightstats/hub/dao/aws/DynamoWebhookDao.java
```diff
@@ -74,13 +76,23 @@ public Webhook get(String name) {
...
             log.warn("group not found {}", name, e);
             return null;
         }
     }
 
+    private Optional<Webhook> tryMapItem(Map<String, AttributeValue> item) {
+        try {
+            return Optional.of(mapItem(item));
+        } catch (Exception e) {
+            log.warn("Unable to map webhook config {} {}", item.get("key"), e.getMessage());
```
chriskessel: It's generally an anti-pattern to eat the underlying exception as that loses the stack trace. Usually the log line is:
`log.warn("bad things happened", e)`
Unless there's a specific reason not to? Like maybe this throws exceptions a LOT and it'd flood the logs?

lkemmerer: Aha. I copy-pasta'd this line from the channel config DAO. I'll fix it in both places.

chriskessel: Woohoo, my review found something useful!

lkemmerer: You say this like it's not a common occurrence. ;)

src/main/java/com/flightstats/hub/dao/aws/DynamoWebhookDao.java
```diff
@@ -74,13 +76,23 @@ public Webhook get(String name) {
             if (result.getItem() == null) {
                 return null;
             }
-            return mapItem(result.getItem());
+            return tryMapItem(result.getItem())
+                    .orElseThrow(() -> new ResourceNotFoundException("Unable to read webhook config from dynamo " + name));
```
manjula-chikkanagappa: The flow never hits path orElseThrow(()). tryMapItem returns empty instance or value.
Also in this method there are 3 return statements. can this consolidated to single exit?

src/test/java/com/flightstats/hub/dao/aws/DynamoWebhookDaoTest.java
```diff
@@ -128,4 +163,11 @@ private Webhook buildGenericWebhook(String name) {
         dynamoEntry.put("batch", new AttributeValue(webhook.getBatch()));
         return dynamoEntry;
     }
+
+    private Map<String, AttributeValue> createBogusEntry() {
+        Map<String, AttributeValue> bogusRecord = new HashMap<>();
+        bogusRecord.put("i exist", new AttributeValue("to break your code"));
```
manjula-chikkanagappa: funny :)

src/main/java/com/flightstats/hub/dao/aws/DynamoWebhookDao.java
lkemmerer: I did some refactoring to reduce the number of returns and to remove the confusing `orElseThrow` bit.

src/main/java/com/flightstats/hub/dao/aws/DynamoWebhookDao.java
```diff
@@ -69,31 +68,34 @@ public void upsert(Webhook webhook) {
...
         try {
             GetItemResult result = dbClient.getItem(dynamoProperties.getWebhookConfigTableName(), keyMap, true);
-            if (result.getItem() == null) {
-                return null;
-            }
-            return tryMapItem(result.getItem())
-                    .orElseThrow(() -> new ResourceNotFoundException("Unable to read webhook config from dynamo " + name));
+            return Optional.ofNullable(result)
+                    .map(GetItemResult::getItem);
         } catch (ResourceNotFoundException e) {
```
manjula-chikkanagappa: Hmm. With this change the problem shifted to next step :) Exception block is never hit when getItem is null. Optional.ofNullable returns Optional.empty when item is null.

manjula-chikkanagappa: Had a offline convo with Laurie. The ResourceNotFoundException is for the dbClient.getItem call. This makes sense now.

lkemmerer: Talked to Manjula about this. Like 80 throws a `ResourceNotFoundException` during `getItem`. Then the AWS SDK API is super helpful by adding a second `getItem` method to get the map of record attributes, which...apparently might be null.

PR #1236, opened 2020-01-28
Clean up some webhook validation tests

lkemmerer:
I'm not sure why tests in this file were only intermittently failing, but I updated the test class to use lenient mocking, because otherwise the tests are sensitive to the order of validations. I couldn't think of a reason that ordering would be particularly important.

While I was in there, I standardized the way we set up the webhook under test. Now it'll build off a valid webhook and only override the fields with which we want to trigger the validation exception.

### DIFF
--- a/src/test/java/com/flightstats/hub/webhook/WebhookValidatorTest.java
+++ b/src/test/java/com/flightstats/hub/webhook/WebhookValidatorTest.java
@@ -3,158 +3,151 @@ package com.flightstats.hub.webhook;
 import com.flightstats.hub.config.properties.AppProperties;
 import com.flightstats.hub.config.properties.WebhookProperties;
 import com.flightstats.hub.exception.InvalidRequestException;
+import com.flightstats.hub.model.WebhookType;
 import com.google.common.base.Strings;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
 
+import java.util.function.Function;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 @ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
 class WebhookValidatorTest {
 
     private static final int CALLBACK_TIMEOUT_DEFAULT_IN_SEC = 120;
-
-    @Mock
-    private AppProperties appProperties;
-    @Mock
-    private WebhookProperties webhookProperties;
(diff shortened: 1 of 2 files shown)


src/test/java/com/flightstats/hub/webhook/WebhookValidatorTest.java
```diff
@@ -3,158 +3,151 @@
...
     @Test
     void testInvalidLocalhost() {
-        webhook = Webhook.builder()
-                .callbackUrl("http:/localhost:8080/url")
+        WebhookValidator webhookValidator = new WebhookValidator(getDefaultAppProperties(), getDefaultWebhookProperties());
+        Webhook webhook = getWebhook(builder -> builder.callbackUrl("http://localhost:8080/url"));
+        assertThrows(InvalidRequestException.class, () -> webhookValidator.validate(webhook));
+    }
+
+    private Webhook getWebhook(Function<Webhook.WebhookBuilder, Webhook.WebhookBuilder> customizations) {
```
lkemmerer: This uses a `Function<WebhookBuilder, WebhookBuilder>` rather than using `with`ers (i.e. `getWebhook().withName("invalid N@ame")`) because there are a few fields that can't be overridden once the webhook is built. I wanted creation of the webhooks to be consistent even with those fields, and this solved that problem.

manjula-chikkanagappa: How is function call different from building webhook directly? Like instead of  line 138   customizations.apply(webhookBuilder).build(); the call could end on line 137 .heartbeat(false).build()
How are they different?

src/test/java/com/flightstats/hub/webhook/WebhookValidatorTest.java
```diff
@@ -3,158 +3,151 @@
...
-
-    @BeforeEach
-    void setUp() {
-        webhookValidator = new WebhookValidator(appProperties, webhookProperties);
-        webhook = Webhook.builder()
-                .callbackUrl("http://client/url")
-                .channelUrl("http://hub/channel/channelName")
-                .parallelCalls(1)
-                .build();
-    }
```
manjula-chikkanagappa: Thanks for cleaning up redundant code and this test looks lot cleaner now.

src/test/java/com/flightstats/hub/webhook/WebhookValidatorTest.java
lkemmerer: Hopefully I'm answering the right question here... There are a few fields that can't be changed after you call `build()`; I think those fields are `name`, `channelUrl`, and `callbackUrl`.
So if line 137 did `.heartbeat(false).build()`, I couldn't use `getWebhook().withName("invalid N@me")` in the test because there's no wither for name. The old implementation of tests for those fields used `Webhook.builder().name(...)...build` in the test to get around that.

Unfortunately, there was at least one test for channel url that was failing because one of the OTHER fields specified in the build statement wasn't valid. Using a function lets the test inject the new, invalid values for those fields while maintaining a single definition of what a valid template for a webhook looks like...and kinda makes it more obvious about which fields/values are important to the test.

Sorry for writing so much there. Does that answer your question?

manjula-chikkanagappa: That explains it. Thanks for the write up :)

PR #1237, opened 2020-01-28
Use lenient mocking for the statsd filter test.

lkemmerer:
We have tests against some boolean functions with mocked values that would short-circuit before evaluating the second part. We don't care which side is checked first, so the tests mock out both sides.
This PR adds leniency to the mocking so we don't have to code up the tests in a way that implicitly indicates that order matters by only mocking the first side.

btw, I think something's weird with our test setup, because these tests don't fail consistently. This is already at least one level of yak shaving more than I'd like to do, so I'm going to put off trying to figure out why.

### DIFF
--- a/src/test/java/com/flightstats/hub/metrics/StatsDFilterTest.java
+++ b/src/test/java/com/flightstats/hub/metrics/StatsDFilterTest.java
@@ -12,6 +12,8 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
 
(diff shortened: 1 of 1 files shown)


src/test/java/com/flightstats/hub/metrics/StatsDFilterTest.java
```diff
@@ -21,6 +23,7 @@
 import static org.mockito.Mockito.when;
 
 @ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
 class StatsDFilterTest {
```
manjula-chikkanagappa: Since this setting is required only for some tests, it would be nice to add at method level
Mockito.lenient().when(mock.foo(1)).thenReturn(1);

lkemmerer: Cool, I didn't know about that (and didn't look too closely because I really just wanted to stabilize the build). I'll swap it out.

PR #1239, opened 2020-02-04
Rename some webhook classes

lkemmerer:
We've got a million `Webhook` classes and a million `Manager` classes. This tries to name some of them more aptly.
- `WebhookManager` is the class that's responsible for making sure that all active webhooks are running on one and only one node. `Coordinator` seemed slightly more appropriate, even if it's also an overused name.
- `LocalWebhookManager` is where webhooks are actually running, updating state, delivering items, etc. It's who we ask if we want to know what webhooks a particular node is actually leading. I dropped `Manager` for the still bad, but slightly more specific `Runner`.
- `WebhookContentPathSet` aka groupSet aka groupInFlight is now `WebhookContentInFlight` and it's ZK state of all the content keys that are currently trying to be delivered to webhook callbacks.

I'm open to better names. These made it easier for me to reason about flow for the data loss bug...

### DIFF
--- a/src/test/java/com/flightstats/hub/webhook/WebhookCoordinatorTest.java
+++ b/src/test/java/com/flightstats/hub/webhook/WebhookCoordinatorTest.java
@@ -39,9 +39,9 @@ import static org.mockito.MockitoAnnotations.initMocks;
 @ExtendWith({MockitoExtension.class})
 @MockitoSettings(strictness = Strictness.LENIENT)
 @Execution(ExecutionMode.SAME_THREAD)
-class WebhookManagerTest {
+class WebhookCoordinatorTest {
     @Mock
-    private LocalWebhookManager localWebhookManager;
+    private LocalWebhookRunner localWebhookRunner;
     @Mock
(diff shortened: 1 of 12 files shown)


src/test/java/com/flightstats/hub/webhook/WebhookCoordinatorTest.java
```diff
@@ -201,11 +201,11 @@ void testIfNotLeaderWontRegisterServices() {
         services.forEach((type, svcs) -> assertTrue(type + " has services registered", svcs.isEmpty()));
     }
 
-    private WebhookManager getWebhookManager() {
-        return new WebhookManager(
-                localWebhookManager,
+    private WebhookCoordinator getWebhookManager() {
```
Paul-Hess: Does it make sense to rename this getter to getWebhookCoordinator?

lkemmerer: Yep, will do!

PR #1240, opened 2020-02-06
Fix a data loss bug related to restarting nodes and inappropriately deleting state in ZK

lkemmerer:
This fixes a bug where, under certain restart + race-condition circumstances, all webhook state for a webhook in zookeeper was being destroyed when a webhook was stopped on a node. This caused some batch channel data loss when we did planned hub downtime -- and possibly, maybe during other times. It's hard to know.

The fix is basically to ensure that the only webhook-related data changed in Zookeeper when a node stops a webhook is that the node is no longer running the webhook. All other state should remain (i.e. the last completed info, which items are currently being delivered, and what items have errored out).

While I was researching this bug, I realized that we mix up and misuse some verbs related to webhook actions. Those actions in particular are "delete," "stop," and "remove."  I changed a bunch of names to help align with the difference. "Stop" now refers to stopping a webhook from running on a node. "Delete" refers to both deleting webhook state info in Zookeeper and actually deleting the webhook configuration from dynamo, which I think always go hand-in-hand.

Behavior around stopping and deleting should look something like this:
- Turning off a node should STOP a webhook and leave all other state as-is.
- The following should STOP the webhook from running, DELETE the webhook state in ZK, and then DELETE the webhook configuration in dynamo:
  - Deleting a webhook
  - Updating a webhook cursor (which should then recreate and restart with the new latest item).
  - Changing a BOTH channel to SINGLE, which deletes the internal batch webhook we use to zip items and ship them to S3.
  - Changing a channel that uses replication to not use replication, which deletes the intenral replication webhook.
  - Deleting a BATCH or BOTH channel OR a replicated channel, which deletes the related internal webhook.

That's a lot of words. Hopefully they make sense to you and future me. Please reach out if you think talking IRL about the changes would be helpful!

### DIFF
--- a/src/main/java/com/flightstats/hub/util/HubUtils.java
+++ b/src/main/java/com/flightstats/hub/util/HubUtils.java
@@ -116,7 +116,7 @@ public class HubUtils {
         }
     }
 
-    public void stopGroupCallback(String groupName, String sourceChannel) {
+    public void deleteWebhook(String groupName, String sourceChannel) {
         String groupUrl = getSourceUrl(sourceChannel) + "/group/" + groupName;
         log.debug("stopping {} ", groupUrl);
         ClientResponse response = null;
@@ -125,7 +125,7 @@ public class HubUtils {
                     .accept(MediaType.APPLICATION_JSON)
                     .type(MediaType.APPLICATION_JSON)
                     .delete(ClientResponse.class);
-            log.debug("stop group response {}", response);
+            log.debug("delete webhook {}", response);
         } finally {
             HubUtils.close(response);
         }
--- a/src/main/java/com/flightstats/hub/util/HubUtils.java
+++ b/src/main/java/com/flightstats/hub/util/HubUtils.java
@@ -117,11 +117,11 @@ public class HubUtils {
     }
 
     public void deleteWebhook(String groupName, String sourceChannel) {
-        String groupUrl = getSourceUrl(sourceChannel) + "/group/" + groupName;
-        log.debug("stopping {} ", groupUrl);
+        String webhookUrl = getSourceUrl(sourceChannel) + "/group/" + groupName;
+        log.debug("deleting {} ", webhookUrl);
         ClientResponse response = null;
         try {
-            response = followClient.resource(groupUrl)
+            response = followClient.resource(webhookUrl)
                     .accept(MediaType.APPLICATION_JSON)
                     .type(MediaType.APPLICATION_JSON)
                     .delete(ClientResponse.class);
--- a/src/main/java/com/flightstats/hub/webhook/WebhookLeader.java
+++ b/src/main/java/com/flightstats/hub/webhook/WebhookLeader.java
@@ -157,9 +157,6 @@ class WebhookLeader implements Lockable {
             closeStrategy();
             stopExecutor();
             webhookStateReaper.stop(webhook.getName());
-            if (deleteOnExit.get()) {
-                webhookStateReaper.delete(webhook.getName());
-            }
             log.info("stopped last completed at {} {}", webhookStrategy.getLastCompleted(), webhook.getName());
             statsdReporter.incrementEventCompletion(LEADERSHIP_METRIC, "name:" + webhook.getName());
             webhookStrategy = null;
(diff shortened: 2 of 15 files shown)


src/test/java/com/flightstats/hub/webhook/WebhookStateReaperTest.java
```diff
@@ -179,11 +188,121 @@ void testDoesNothingIfLeadershipDisabled() throws Exception {
...
         assertWebhookInProcessExists(webhookName);
         assertWebhookLeaderExists(webhookName);
+    }
 
-        // ...AND then clean up the state so the next test can run.  Eww.
-        when(webhookProperties.isWebhookLeadershipEnabled()).thenReturn(true);
-        reaper = new WebhookStateReaper(clusterCacheDao, contentKeysInFlight, webhookErrorService, webhookLeaderLocks, webhookProperties);
-        reaper.delete(webhookName);
+    @Test
+    void testStopCleansUpZookeeperNodesRelatedToLeadershipStateButNotProgressOrErrors() throws Exception {
```
manjula-chikkanagappa: Oooh! Thats a very loooooooooooong test name! Since this test method does not need to be invoked manually in our code, its should be fine. It would be nice to break it up with some _ for easy reading though. Something like
testStop_cleansUpZookeeperNodeLeaderState_notProgressOrErrors

lkemmerer: sounds good! I'll do that for all the tests in this file.

src/main/java/com/flightstats/hub/util/HubUtils.java
```diff
@@ -116,7 +116,7 @@ public void startWebhook(Webhook webhook) {
         }
     }
 
-    public void stopGroupCallback(String groupName, String sourceChannel) {
+    public void deleteWebhook(String groupName, String sourceChannel) {
         String groupUrl = getSourceUrl(sourceChannel) + "/group/" + groupName;
         log.debug("stopping {} ", groupUrl);
```
Paul-Hess: It looks like to me that this logger should be updated from "stopping" to "deleting" as well

src/main/java/com/flightstats/hub/webhook/WebhookLeader.java
```diff
@@ -262,12 +259,8 @@ private void send(ContentPath contentPath) throws InterruptedException {
                 long start = System.currentTimeMillis();
                 boolean shouldGoToNextItem = retryer.send(webhook, contentPath, webhookStrategy.createResponse(contentPath));
                 statsdReporter.time("webhook", start, "name:" + webhook.getName());
-                if (shouldGoToNextItem) {
-                    if (increaseLastUpdated(contentPath)) {
-                        if (!deleteOnExit.get()) {
-                            clusterCacheDao.setIfNewer(contentPath, webhook.getName(), WEBHOOK_LAST_COMPLETED);
-                        }
-                    }
+                if (shouldGoToNextItem && increaseLastUpdated(contentPath)) {
```
Paul-Hess: This would read easier for me if "increaseLastUpdated" was name explicitly that it only conditionally updates last updated, which is implied because it returns a boolean, but still IMO, sorry, that name obviously predates your changes

src/main/java/com/flightstats/hub/webhook/WebhookLeader.java
```diff
@@ -262,12 +259,8 @@ private void send(ContentPath contentPath) throws InterruptedException {
...
+                    clusterCacheDao.setIfNewer(contentPath, webhook.getName(), WEBHOOK_LAST_COMPLETED);
                 }
                 keysInFlight.remove(webhook.getName(), contentPath);
```
Paul-Hess: Just a consideration, I am probably not grokking all going on here but it looks to me logically like this should move inside of the if block?

lkemmerer: ...I have no idea. I'll see if I can make any sense of it. May leave it for now so as to not change behavior.

src/main/java/com/flightstats/hub/webhook/WebhookLeader.java
lkemmerer: That's a small change and makes sense to do. 👍

src/main/java/com/flightstats/hub/webhook/WebhookLeader.java
lkemmerer: I looked into this a little more and keysInFlight tracks the stuff that's being sent *right now*. It looks like it has at least one purpose: making sure that every item that's being delivered finishes prior to turning off the webhook.
So having it outside the if block might make sense? I think. There are no unit-type tests around this, I think, so I'm going to leave it alone.

PR #1242, opened 2020-02-20
Timed webhook strategy cleanup

lkemmerer:
There's no rush on getting feedback on this, as the team has other priorities.

This is refactoring and additional tests that were done while I was trying to track down lost minutes of batched data. I didn't uncover any bugs with the code while I was trying to add tests, but the code might be mildly easier to understand with some of this work. Hopefully.

I think it might be easiest to understand if you read the code commit-by-commit and in split diff mode rather than unified. I'm going to add some comments to help explain what stuff's doing and where things got moved.

This code is pretty deep down in hub functionality, so if it'd help to have an ad-hoc meeting about what this code does and what changed, lemme know.

### DIFF
--- a/src/main/java/com/flightstats/hub/webhook/WebhookCoordinator.java
+++ b/src/main/java/com/flightstats/hub/webhook/WebhookCoordinator.java
@@ -7,6 +7,7 @@ import com.flightstats.hub.cluster.Watcher;
 import com.flightstats.hub.config.properties.WebhookProperties;
 import com.flightstats.hub.dao.Dao;
 import com.flightstats.hub.model.ContentPath;
+import com.flightstats.hub.webhook.strategy.WebhookStrategy;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.AbstractIdleService;
 import com.google.common.util.concurrent.AbstractScheduledService;
--- a/src/main/java/com/flightstats/hub/webhook/strategy/TimedWebhookStrategy.java
+++ b/src/main/java/com/flightstats/hub/webhook/strategy/TimedWebhookStrategy.java
@@ -1,4 +1,4 @@
-package com.flightstats.hub.webhook;
+package com.flightstats.hub.webhook.strategy;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
@@ -16,6 +16,7 @@ import com.flightstats.hub.model.SecondPath;
 import com.flightstats.hub.model.TimeQuery;
 import com.flightstats.hub.util.RuntimeInterruptedException;
 import com.flightstats.hub.util.TimeUtil;
+import com.flightstats.hub.webhook.Webhook;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import lombok.extern.slf4j.Slf4j;
 import org.joda.time.DateTime;
--- a/src/main/java/com/flightstats/hub/webhook/strategy/TimedWebhookStrategy.java
+++ b/src/main/java/com/flightstats/hub/webhook/strategy/TimedWebhookStrategy.java
@@ -17,6 +17,7 @@ import com.flightstats.hub.model.TimeQuery;
 import com.flightstats.hub.util.RuntimeInterruptedException;
 import com.flightstats.hub.util.TimeUtil;
 import com.flightstats.hub.webhook.Webhook;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import lombok.extern.slf4j.Slf4j;
 import org.joda.time.DateTime;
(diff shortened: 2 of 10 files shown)


src/main/java/com/flightstats/hub/webhook/strategy/TimedWebhookStrategy.java
```diff
@@ -234,8 +234,19 @@ private DateTime getNextTime() {
                 }
                 return getNextTime.apply(lastTime);
             }
+        };
+    }
 
-        }, getOffsetSeconds.get(), period, TimeUnit.SECONDS);
+    @Override
+    public void start(Webhook webhook, ContentPath initialStartingPath) {
+        ThreadFactory factory = new ThreadFactoryBuilder().setNameFormat(webhook.getBatch() + "-webhook-" + webhook.getName() + "-%s").build();
```
lucasmo: I know this isn't your code, but if this webhook is started multiple times the threads will have identical names. Not sure how possible that is.

lkemmerer: It *shouldn't* happen. I'm not sure how bomb-proof that logic is, but I can check.

src/main/java/com/flightstats/hub/webhook/strategy/TimedWebhookStrategy.java
```diff
@@ -157,15 +163,9 @@ public ContentPath getLastCompleted() {
...
-    public void start(Webhook webhook, ContentPath startingPath) {
-        ThreadFactory factory = new ThreadFactoryBuilder().setNameFormat(webhook.getBatch() + "-webhook-" + webhook.getName() + "-%s").build();
-        executorService = Executors.newSingleThreadScheduledExecutor(factory);
-        log.info("starting {} with starting path {}", webhook, startingPath);
-        executorService.scheduleAtFixedRate(new Runnable() {
-
-            ContentPath lastAdded = startingPath;
-
+    @VisibleForTesting
+    Runnable getContentKeyGenerator(Webhook webhook, ContentPath initialStartingPath) {
```
lkemmerer: This was extracted from the start method so it could be tested without messing with threads.

src/main/java/com/flightstats/hub/webhook/strategy/TimedWebhookStrategy.java
```diff
@@ -184,43 +184,69 @@ public void run() {
...
                 }
             }
+            ContentPath lastAdded = initialStartingPath;
 
             private void doWork() throws InterruptedException {
-                DateTime nextTime = getNextTime.apply(lastAdded.getTime());
-                if (lastAdded instanceof ContentKey) {
-                    nextTime = lastAdded.getTime();
-                }
-                DateTime stable = TimeUtil.stable().minus(duration);
```
lkemmerer: This through line 201 was extracted into a separate method.

src/main/java/com/flightstats/hub/webhook/strategy/TimedWebhookStrategy.java
```diff
@@ -244,6 +270,13 @@ private void doWork() throws InterruptedException {
         return Optional.ofNullable(queue.poll(10, TimeUnit.MINUTES));
     }
 
+    @VisibleForTesting
+    SortedSet<ContentPathKeys> drainQueue() {
```
lkemmerer: This allows the test to interrogate what was added to the queue after running `getContentKeyGenerator`

src/main/java/com/flightstats/hub/webhook/WebhookCoordinator.java
```diff
@@ -7,6 +7,7 @@
 import com.flightstats.hub.config.properties.WebhookProperties;
 import com.flightstats.hub.dao.Dao;
 import com.flightstats.hub.model.ContentPath;
+import com.flightstats.hub.webhook.strategy.WebhookStrategy;
 import com.google.common.annotations.VisibleForTesting;
```
Paul-Hess: this and the next two changed files look a little odd because there is a new import, but not a change as to the import's use in the file, just checking that these are intentional

lkemmerer: Hopefully I understand your question correctly: When I moved the webhook strategy stuff into its own package, it required me to import them in all the classes that used it and with which it used to live (so anything in just "com.flightstats.hub.webhook"). That's why there's no code changed: the file(s) just moved.

src/main/java/com/flightstats/hub/webhook/strategy/TimedWebhookStrategy.java
lkemmerer: I moved this up into the constructor (so that it'll increment properly if `start` is called multiple times) and also added 4 characters of randomness in case a `TimedWebhookStrategy` is initialized multiple times for the same webhook. Might be overkill; the 4 characters of randomness is likely enough...

PR #1248, opened 2020-04-02
Don't try to start paused webhooks when ensuring that webhooks are running properly

lkemmerer:
This isn't quite ready for review as I need to write up a description and maybe add some comments (and do one last run through CI). I'll add reviewers when it's ready.

### DIFF
--- a/src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
+++ b/src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
@@ -10,12 +10,12 @@ import lombok.extern.slf4j.Slf4j;
 
 import javax.inject.Inject;
 import javax.inject.Named;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import static com.flightstats.hub.constant.NamedBinding.WEBHOOK_CLIENT;
 import static java.util.stream.Collectors.toList;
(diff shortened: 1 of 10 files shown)


src/main/java/com/flightstats/hub/webhook/WebhookLeaderState.java
```diff
@@ -28,13 +31,40 @@ public ActiveWebhooks(
...
+    public static class RunningState {
+        boolean leadershipAcquired;
+        Set<String> runningServers;
+
+        public boolean isRunningOnSingleServer() {
+            return isLeadershipAcquired() && getRunningServers().size() == 1;
+        }
+
+        public boolean isStopped() {
+            return !isLeadershipAcquired() && getRunningServers().isEmpty();
```
chriskessel, line 63: Depending on the construction, runningServers could be null. You could add `@Builder.Default` to the instance declaration.
```
@Builder.Default
Set<String> runningServers() = Collections.emptySet();
```

src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
```diff
@@ -45,22 +44,45 @@ boolean stop(String name, String server) {
         return put(server + "/internal/webhook/stop/" + name);
     }
 
-    Optional<String> runOnServerWithFewestWebhooks(String name) {
-        return runOnOneServer(name, getOrderedServers());
+    Optional<String> runOnOnlyOneServer(String name, Collection<String> runningServers) {
```
lkemmerer, line 47: This new method does the following:
1.Tries to confirm the webhook is running/started on one of the nodes in the list
1. Sends stop requests to the remaining nodes that were passed in. 
1. If none of the nodes in the list responded that the run is successful, it'll try to fallback to one of the remaining nodes in the cluster that was not in the list.

src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
```diff
@@ -45,22 +44,45 @@ boolean stop(String name, String server) {
...
+        return runOnServerWithFewestWebhooksExcluding(name, Collections.emptyList());
+    }
+
+    private Optional<String> runOnOneServer(String name, Stream<String> servers) {
+        return servers
                 .filter(server -> put(server + "/internal/webhook/run/" + name))
                 .findFirst();
     }
 
+    private Optional<String> runOnServerWithFewestWebhooksExcluding(String name, Collection<String> servers) {
```
lkemmerer, line 75: I feel like this class has gotten a little out of hand. Maybe there's a way to clean it up, or separate it so there's less going on. ¯\_(ツ)_/¯

lkemmerer:
Going to close this so that I can give some of the commits clearer commit messages, which'll require a force push.

PR #1249, opened 2020-04-02
HUB-994: Don't try to start a paused webhook in our periodically-run webhook coordinator

lkemmerer:
The bug fix (in the last commit) addresses the following scenario:
1. A user pauses a webhook
2. The WebhookCoordinator scheduled task runs, sees that the paused webhook is not currently running, and sends off an internal request to start it up. 
3. Somewhere down the stack in that internal start resource call, the run request fails (because the webhook is paused) and returns a 400 error.
4. This happens on each node every x (x=5?) minutes. When I paused all the batch webhooks, this caused havoc on the cluster.

It fixes it by, y'know, taking webhook paused state into consideration when the `WebhookCoordinator` decides what actions to take re: running the webhook.

Commits 2, 4, 6, and 7 contain actual refactoring and changes to behavior. Those commit messages include the main class to look at and the responsibility of that class.

### DIFF
--- a/src/main/java/com/flightstats/hub/dao/aws/S3BatchManager.java
+++ b/src/main/java/com/flightstats/hub/dao/aws/S3BatchManager.java
@@ -65,8 +65,9 @@ public class S3BatchManager {
                     appProperties.getAppEnv());
 
             if (channel.isSingle()) {
-                if (!webhookLeaderState.getServers(channel.getName()).isEmpty()) {
-                    log.debug("turning off batch webhook {}", channel.getDisplayName());
+                WebhookLeaderState.RunningState state = webhookLeaderState.getState(channel.getName());
+                if (!state.isStopped()) {
+                    log.debug("turning off batch webhook for {}", channel.getDisplayName());
                     s3Batch.delete();
                 }
             } else {
--- a/src/main/java/com/flightstats/hub/dao/aws/S3BatchManager.java
+++ b/src/main/java/com/flightstats/hub/dao/aws/S3BatchManager.java
@@ -7,7 +7,7 @@ import com.flightstats.hub.dao.Dao;
 import com.flightstats.hub.model.ChannelConfig;
 import com.flightstats.hub.replication.S3Batch;
 import com.flightstats.hub.util.HubUtils;
-import com.flightstats.hub.webhook.ActiveWebhooks;
+import com.flightstats.hub.webhook.WebhookLeaderState;
 import com.flightstats.hub.webhook.Webhook;
 import com.flightstats.hub.webhook.WebhookService;
 import com.google.common.util.concurrent.AbstractIdleService;
@@ -27,20 +27,20 @@ public class S3BatchManager {
     private final WebhookService webhookService;
     private final Dao<ChannelConfig> channelConfigDao;
     private final HubUtils hubUtils;
-    private final ActiveWebhooks activeWebhooks;
+    private final WebhookLeaderState webhookLeaderState;
     private final AppProperties appProperties;
 
     @Inject
     public S3BatchManager(WebhookService webhookService,
                           @Named("ChannelConfig") Dao<ChannelConfig> channelConfigDao,
                           HubUtils hubUtils,
-                          ActiveWebhooks activeWebhooks,
+                          WebhookLeaderState webhookLeaderState,
                           AppProperties appProperties,
                           S3Properties s3Properties) {
         this.webhookService = webhookService;
         this.channelConfigDao = channelConfigDao;
         this.hubUtils = hubUtils;
-        this.activeWebhooks = activeWebhooks;
+        this.webhookLeaderState = webhookLeaderState;
         this.appProperties = appProperties;
 
         if (s3Properties.isBatchManagementEnabled()) {
@@ -65,7 +65,7 @@ public class S3BatchManager {
                     appProperties.getAppEnv());
 
             if (channel.isSingle()) {
-                if (!activeWebhooks.getServers(channel.getName()).isEmpty()) {
+                if (!webhookLeaderState.getServers(channel.getName()).isEmpty()) {
                     log.debug("turning off batch webhook {}", channel.getDisplayName());
                     s3Batch.delete();
                 }
--- a/src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
+++ b/src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
@@ -10,12 +10,12 @@ import lombok.extern.slf4j.Slf4j;
 
 import javax.inject.Inject;
 import javax.inject.Named;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import static com.flightstats.hub.constant.NamedBinding.WEBHOOK_CLIENT;
 import static java.util.stream.Collectors.toList;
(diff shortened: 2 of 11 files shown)


src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
```diff
@@ -45,16 +45,40 @@ boolean stop(String name, String server) {
         return put(server + "/internal/webhook/stop/" + name);
     }
 
-    Optional<String> runOnServerWithFewestWebhooks(String name) {
-        return runOnOneServer(name, getOrderedServers());
+    Optional<String> runOnOnlyOneServer(String name, Collection<String> runningServers) {
```
lkemmerer, line 47: This new method does the following:

1. Tries to confirm the webhook is running/started on one of the nodes in the list
2. Sends stop requests to the remaining nodes that were passed in.
3. If none of the nodes in the list responded that the run is successful, it'll try to fallback to one of the remaining nodes in the cluster that was not in the list.

src/main/java/com/flightstats/hub/dao/aws/S3BatchManager.java
```diff
@@ -65,8 +65,9 @@ private void setupBatch() {
                     appProperties.getAppEnv());
 
             if (channel.isSingle()) {
-                if (!webhookLeaderState.getServers(channel.getName()).isEmpty()) {
-                    log.debug("turning off batch webhook {}", channel.getDisplayName());
+                WebhookLeaderState.RunningState state = webhookLeaderState.getState(channel.getName());
+                if (!state.isStopped()) {
+                    log.debug("turning off batch webhook for {}", channel.getDisplayName());
```
Paul-Hess, line 70: why is there a s3batch.delete if the channel.isSingle?

Paul-Hess: this is unrelated to your changes, I'm just not sure what is going on here

lkemmerer: I believe it's checking to see if the channel changed from batch (or both?) to single storage mode (single channel, but the batch webhook is still running) and stops the webhook if it's in that state.

lkemmerer: that may not be entirely accurate. I knew what this did at one point, I think.

src/main/java/com/flightstats/hub/dao/aws/S3BatchManager.java
```diff
@@ -65,8 +65,9 @@ private void setupBatch() {
                     appProperties.getAppEnv());
 
             if (channel.isSingle()) {
-                if (!webhookLeaderState.getServers(channel.getName()).isEmpty()) {
-                    log.debug("turning off batch webhook {}", channel.getDisplayName());
+                WebhookLeaderState.RunningState state = webhookLeaderState.getState(channel.getName());
+                if (!state.isStopped()) {
```
manjula-chikkanagappa, line 69: Should this be paused? Stopped and Paused are used interchangeably here. You seem to use pause in your comments and stopped in actual code.

src/test/java/com/flightstats/hub/webhook/WebhookCoordinatorTest.java
```diff
@@ -106,7 +112,11 @@ void testWhenNewWebhook_getsAddedToServerManagingFewestWebhooks() {
 
     @Test
     void testWhenInactiveWebhook_getsAddedToServerManagingFewestWebhooks() {
-        when(webhookLeaderState.isActiveWebhook(WEBHOOK_NAME)).thenReturn(false);
+        when(webhookLeaderState.getState(WEBHOOK_NAME)).thenReturn(
+                WebhookLeaderState.RunningState.builder()
+                        .leadershipAcquired(false)
+                        .runningServers(newHashSet())
+                        .build());
```
manjula-chikkanagappa, line 108: Code will be easier to read if this "WebhookLeaderState.RunningState.builder()
                        .leadershipAcquired(false)
                        .runningServers(newHashSet())
                        .build()"
is pulled in to separate method. It is redundant.

src/main/java/com/flightstats/hub/dao/aws/S3BatchManager.java
lkemmerer: `paused` is the name of an attribute in the webhook configuration record. `stopped` is any webhook that is not running, whether it should be paused or not.

src/main/java/com/flightstats/hub/dao/aws/S3BatchManager.java
```diff
@@ -27,20 +27,20 @@
     private final WebhookService webhookService;
     private final Dao<ChannelConfig> channelConfigDao;
     private final HubUtils hubUtils;
-    private final ActiveWebhooks activeWebhooks;
+    private final WebhookLeaderState webhookLeaderState;
```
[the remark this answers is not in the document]
lkemmerer: `ActiveWebhooks` also returns information about not-active webhooks. I'd previously introduced `State` in class names as a pattern that denotes that the class stores or retrieves distributed state in zookeeper, so this brings the class name in line with that. The zookeeper nodes that are dealt with in the class are around the webhook leadership/locking behavior that's mainly used in the `WebhookLeader` class. 
Honestly, I always found `ActiveWebhooks` a pretty confusing name for what it was doing. It's slowly gotten less confusing, but was still a bit misleading.

src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
```diff
@@ -45,16 +45,40 @@ boolean stop(String name, String server) {
...
+                        .map(successfulRun -> !server.equals(successfulRun))
+                        .orElse(true))
+                .forEach(server -> stop(name, server));
+
+        return successfulServerRun;
     }
 
     Optional<String> runOnOneServer(String name, Collection<String> servers) {
-        return servers.stream()
+        return runOnOneServer(name, servers.stream());
```
manjula-chikkanagappa, line 62: This method name is same as another private method. Initially i thought this is recursive function and looking for terminating condition. Is it possible to rename one of them?

src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
```diff
@@ -83,8 +82,7 @@ boolean stop(String name, String server) {
      * We want this to return this list in order from fewest to most
      * If we get a non-200, return max int so it's tried last
      */
-    @VisibleForTesting
```
manjula-chikkanagappa, line 62: Nice. I don't like to use this annotation. It just always means a bad design!

src/main/java/com/flightstats/hub/webhook/WebhookCoordinator.java
```diff
@@ -102,26 +101,45 @@ void ensureRunningOnOnlyOneServer(Webhook daoWebhook, boolean webhookChanged) {
...
+            log.debug("found existing webhook {} running on {}", name, state.getRunningServers());
+            webhookClient.runOnOnlyOneServer(name, state.getRunningServers());
+        } else {
+            log.warn("the webhook coordinator seems to have a bug; this statement shouldn't be reached");
+        }
+
+    }
+
+    @VisibleForTesting
+    static class WebhookActionDirector {
```
manjula-chikkanagappa, line 127: This class has package visibility. Do you still need @VisibleForTesting?

src/main/java/com/flightstats/hub/dao/aws/S3BatchManager.java
lkemmerer: They both have state about last delivered, which items have errored, etc. Paused is just one type of stopped webhook, but a webhook could be stopped for other reasons (a hub node is shutting down and releases its lock, a webhook is in the midst of being deleted, etc.)

src/main/java/com/flightstats/hub/webhook/InternalWebhookClient.java
lkemmerer: One takes a collection and one takes a stream, the behavior is otherwise exactly the same, which seems like it'd be a reasonable candidate for argument overloading... If you feel strongly otherwise, I'll see about changing it.

src/main/java/com/flightstats/hub/webhook/WebhookCoordinator.java
lkemmerer: Oh, no, I believe I can get rid of it. Thanks!

lkemmerer:
By log statements, where do you mean, exactly? Each log statement includes the thread it's running in, which I've found mostly to be sufficient in debugging issues. 
(Also, depending on your answer, I'd rather not have request ids permeate outside the resource or client classes as that seems like it'd break encapsulation.)

src/main/java/com/flightstats/hub/webhook/WebhookCoordinator.java
```diff
@@ -123,8 +123,7 @@ void ensureRunningOnOnlyOneServer(Webhook daoWebhook, boolean webhookChanged) {
 
     }
 
-    @VisibleForTesting
-    static class WebhookActionDirector {
+    private static class WebhookActionDirector {
```
manjula-chikkanagappa, line 126: Does this class have unit tests? If not, is it possible to add some?

PR #1269, opened 2021-08-28
Spoke read fix 2

Paul-Hess:


src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
```diff
@@ -40,47 +51,60 @@ public FileSpokeStore(String spokePath, int spokeTtlMinutes) {
...
     public boolean insert(String path, byte[] payload) {
         return insert(path, new ByteArrayInputStream(payload));
     }
 
+    @SneakyThrows
     public boolean insert(String path, InputStream input) {
         File file = spokeFilePathPart(path);
-        log.trace("insert {} {} {}", file, file.getParentFile().mkdirs(), file.canExecute());
-        try (FileOutputStream output = new FileOutputStream(file)) {
+        File tmpFile = spokeFilePathPart(SPOKE_TMP_PATH + path);
```
lucasmo, line 63: If I'm understanding this correctly, let's say the path was `ChannelName/2021/08/10/01/02/03/456/abcdef`. 

This means that we'll make `.tmp.spokeChannelName/2021/08/10/01/02/03/456/abcdef`. Is that your intent? I would be concerned that this makes double the number of directory inodes.

If we used `File tmpFile = spokeFilePathPart(path + SPOKE_TMP_PATH);` (maybe SPOKE_TMP_SUFFIX?)
we'd make `ChannelName/2021/08/10/01/02/03/456/abcdef.tmp.spoke`

lucasmo: ... although that might have side effects for anything that lists the directory.

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
```diff
@@ -40,47 +51,60 @@ public FileSpokeStore(String spokePath, int spokeTtlMinutes) {
...
+        log.trace("insert {}", file);
+        filesArtificiallyLocked.add(path);
+        try (FileOutputStream output = new FileOutputStream(tmpFile)) {
             long copy = ByteStreams.copy(input, output);
-            boolean setExecutable = file.setExecutable(true);
-            log.trace("copied {} {} {}", file, copy, setExecutable);
+            log.trace("copied {} {}", file, copy);
+        } catch (IOException e) {
+            log.error("Error writing to spoke path (tmp file phase) {}", tmpFile.getPath(), e);
+            return false;
```
lucasmo, line 72: In the error condition here, `filesArtificiallyLocked` is left populated.

Maybe this structure?
```
try {
   try (output) {
     ...
   }
   move
} catch {
  ...
} finally {
  ...
}
```

lucasmo (approved):
Thanks for getting rid of this gross abuse of the execute bit

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
Paul-Hess: yes, done.

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
Paul-Hess: the full path would be more like /spoke/read_or_write/.tmp.spokeChannelName/.... 
though I think that makes no difference with regards to your point. with SPOKE_TMP_SUFFIX I think we still double the inode usage though only for a matter of (however long it takes to write the file/move it) is there an advantage to file inodes over directory inodes? I added as the suffix you suggested because no matter what I think it looks cleaner


I looked more at the command that looks for the oldest item and either way above would affect it so I added an ignore on the suffix to the find options

PR #1271, opened 2021-09-03
filter out temp suffix from file spoke store queries

Paul-Hess:


src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
```diff
@@ -178,14 +181,19 @@ private void keysInBucket(String key, OutputStream output) {
         }
         try {
             Collection<File> files;
-            if (resolution.equals("second")) {
+            if (Optional.ofNullable(resolution).orElse("FALSE").equals("second")) {
```
lucasmo, line 184: I'm probably too oldschool but is this preferred style? or does `"second".equals(resolution)` solve the NPE? (This is more a question than a request to change this)

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
```diff
@@ -246,6 +258,15 @@ private String getLatest(String channel, String limitPath, DateTime hourToSearch
         return getLatest(channel, limitPath, previous);
     }
 
+    private boolean doesNotEndWith(File file) {
```
lucasmo, line 261: This method name is unclear - `isNotTempFile`?

lucasmo (approved):

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
chriskessel: I was going to say the same thing. I'm not seeing value in the Optional here.

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
```diff
@@ -178,14 +181,19 @@ private void keysInBucket(String key, OutputStream output) {
...
+                files = Arrays.asList(Optional.ofNullable(directory.listFiles(fileFilter)).orElse(new File[]{}));
             } else {
                 files = FileUtils.listFiles(new File(path), null, true);
             }
-            for (File aFile : files) {
+            Collection<File> permanentFiles = files
+                    .stream()
+                    .filter(this::doesNotEndWith)
+                    .collect(Collectors.toList());
+            for (File aFile : permanentFiles) {
```
chriskessel, line 196: Why not use `foreach` after the filter rather than putting it in an list and then looping the list?

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
```diff
@@ -227,7 +235,11 @@ private String getLatest(String channel, String limitPath, DateTime hourToSearch
         log.trace("looking at {} {}", fullHoursPath, minutes);
         for (int i = minutes.length - 1; i >= 0; i--) {
             String minute = minutes[i];
-            String[] fileNames = new File(fullHoursPath + "/" + minute).list();
+            String[] fileNames = Arrays.stream(Optional
+                            .ofNullable(new File(fullHoursPath + "/" + minute).list())
+                            .orElse(new String[]{}))
+                    .filter(this::doesNotEndWith)
+                    .toArray(String[]::new);
             Arrays.sort(fileNames);
```
chriskessel, line 243: I think you can push the sort into the stream.

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
chriskessel: I might go a step further and make it `isTempFile()` and then use negation where that's called. A method with the word "Not" in it is a bit weird.

Or go the other way and name it after whatever is implied if it's not a temp file, `isStandardSpokeFile()` or some such.

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
```diff
@@ -262,20 +283,22 @@ public void getNext(String channel, String startKey, int count, OutputStream out
...
-                            found++;
-                            writeKey(output, channel + "/" + keyFromPath);
-                        }
-                    } else {
+            String[] items = Arrays.stream(
+                            Optional.ofNullable(new File(minute).list())
+                                    .orElse(new String[]{}))
+                    .filter(this::doesNotEndWith)
+                    .toArray(String[]::new);
+            for (String item : items) {
```
chriskessel, line 291: Another case where I think you can use `foreach` on the stream. Unless there's a reason to do it in the for loop that I'm not seeing? I'm not seeing anything modified within in the for loop that would prevent being in a lambda.

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
lkemmerer, line 184: Dumb question, but do we have a constant for "second" that would indicate what its context/meaning is? (My brain needs caffeine and I thought "second" meant the thing after "first" and not -- as one should assume with hub things -- a subdivision of a minute. You can probably ignore this comment.)

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
lkemmerer: If you go this route (which I think you should), i'd extract the writing from to another method (`writeKeyFromFile`?)

lkemmerer, line 196: Unrelated to this line, but related to this method...

The name of this method misled me to think it should have some sort of return value (like the other `keysInBucket` method in this class). Would you be willing to rename this method to `readKeysInBucket`? The calling method with that name seems to just call this method without any sort of additional behavior whatsoever. You'll need to change the access level on this, since I think the other method's not private, though.

lkemmerer (approved):
That `FileSpokeStore` is doing so much work. Nice job figuring out all the touch points.

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
Paul-Hess: done

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
Paul-Hess: done

PR #1273, opened 2021-09-08
Perms issue

Paul-Hess:


docker/Dockerfile
```diff
@@ -11,8 +11,18 @@ ENV MIN_HEAP=1g \
 
 ADD configs /etc/hub
 
+ARG USER_ID
```
lkemmerer, line 14: Do these args need to be added to any scripts since it looks like (at least here) there aren't any default values? I thought we had tests that ran a single node hub, but I might either misremember or not understand what's going on wrt these args.

lkemmerer (approved):

flightstats-it (approved):

PR #1280, opened 2022-02-28
Update JS integration test libraries with vulnerabilities

lkemmerer:
All the tests pass, so I guess we're good? This should resolve some outstanding security vulnerabilities, leaving one in our ruby/docs code.


Paul-Hess:
LGTM

PR #1287, opened 2022-05-18
Fix jmx vuln

Paul-Hess:


lkemmerer:
So this turns off jmx remote...remoting, but still leaves local available. Do we use that feature in any way? If not, are we leaving it turned on because that required fewer changes?

lkemmerer (approved):

PR #1292, opened 2022-07-13
Add in JMX monitoring of Jetty; add ability to adjust thread pool size

lucasmo:
This should allow us to introspect into Jetty's behavior to see if it's behind the request latency. It also allows for tweaking the maximum thread pool size for Jetty, should that be the issue.


lkemmerer (approved):
I deployed this to iad.dev and I can see more beans when I poke around with jmxterm. I'm unsure about which Jetty bean we should take a look at, though.

PR #1296, opened 2022-07-29
add dependabot minimal config to replace / with - in branch names

Paul-Hess:


lkemmerer (approved):
I didn't know we could configure this. Sweet. And thank you!

Paul-Hess:
this will allow us to actually use dependabot PRs with our CI system instead of having to recreate them manually

PR #1299, opened 2022-07-29
Dependabot ruby updates

Paul-Hess:


Paul-Hess:
combines backed-up last 3 ruby dependency updates automated by dependabot, should be easier to merge them individually in future so this will get us caught up

lkemmerer (approved):

PR #1301, opened 2022-08-09
Add tests/refactor request metrics reporting; drop internal channel channel tag

lkemmerer:
This refactors and adds tests. It also removes the channel tag from internal channel-related requests in an attempt to further minimize the number of custom metrics we have.

This temporarily has a log method in place rather than sending data to datadog, as my last dev deploy comically exploded our custom metrics count to something ridiculous and I don't want that to happen again. Once we get some eyes on this, I'll deploy it to a dev node and eyeball it before doing a full dev deploy. Commit-by-commit might be marginally easy. i don't know.


src/main/java/com/flightstats/hub/filter/MetricsRequestFilter.java
```diff
@@ -49,57 +39,42 @@ public void finalStats() {
...
+        long time = System.currentTimeMillis() - startTime;
+        log.trace("request {}, time: {}", metric.getTags().get("endpoint"), time);
+        if (!metric.shouldReport(statsdFilter.getRequestMetricsToIgnore(), statsdFilter::isTestChannel)) {
+            return;
+        }
+
+        String[] tagArray = getTagArray(metric.getTags());
+        log.trace("statsdReporter data sent: {}", Arrays.toString(tagArray));
+        metric.getMetricName().ifPresent(metricName ->
+                statsdReporter.timeTest(metricName, startTime, tagArray));
```
lucasmo, line 62: Is this something temporary that needs to be changed?

src/main/java/com/flightstats/hub/filter/MetricsRequestFilter.java
```diff
@@ -49,57 +39,42 @@ public void finalStats() {
...
         } finally {
             threadLocal.remove();
         }
     }
 
+    @VisibleForTesting
+    void reportTime(RequestMetric metric, long startTime) {
+        long time = System.currentTimeMillis() - startTime;
+        log.trace("request {}, time: {}", metric.getTags().get("endpoint"), time);
+        if (!metric.shouldReport(statsdFilter.getRequestMetricsToIgnore(), statsdFilter::isTestChannel)) {
```
lucasmo, line 55: The `statsdFilter::isTestChannel` logic appears to be in the `StatsdReporter` class itself right before the metric is actually reported, so it can probably be removed here

src/main/java/com/flightstats/hub/metrics/StatsdReporter.java
```diff
@@ -63,6 +65,10 @@ public void time(String name, long start, String... tags) {
         reportWithBothClients(statsDClient -> statsDClient.time(name, System.currentTimeMillis() - start, tags));
     }
 
+    public void timeTest(String name, long start, String... tags) {
```
lucasmo, line 68: As mentioned in the other class, can this be deleted?

src/main/java/com/flightstats/hub/util/HubRequest.java
```diff
@@ -0,0 +1,98 @@
...
+                    .map(UriTemplate::getTemplate)
+                    .map(template -> template.replaceAll(CHARACTERS_TO_REMOVE, ""))
+                    .map(template -> template.replaceAll(CHARACTERS_TO_REPLACE, "_"))
+                    .collect(Collectors.joining(""));
+        }
+
+        public String getChannel() {
+            return Stream.of(getPathParameter("channel"), getHeader("channelName"))
+                    .filter(value -> value != null && !value.isEmpty())
+                    .findFirst().orElse("");
```
lucasmo, line 87: The `.orElse("")` here feels wrong. Should this just return an Optional?

src/main/java/com/flightstats/hub/util/RequestMetric.java
```diff
@@ -0,0 +1,71 @@
...
+
+        return !isMetricToIgnore && !isTestChannel;
+    }
+
+
+    private Optional<String> getChannelTag() {
+        Optional<String> tagAsChannel = request.getTag().map(tag -> "tag/" + tag);
+        return Stream.of(request.getChannel(), tagAsChannel)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
```
lucasmo, line 63: `Optional::stream`? Or is that java 9+ only?

src/main/java/com/flightstats/hub/util/RequestMetric.java
```diff
@@ -0,0 +1,71 @@
...
+            log.info("call to shutdown, ignoring statsd time");
+            return Optional.empty();
+        }
+
+        String requestScope = request.isInternal() ? "internal" : "api";
+        String channelDifferentiator = request.isChannelRelated() ? "channel" : "nonchannel";
+        return Optional.of(String.format("request.%s.%s", requestScope, channelDifferentiator));
+    }
+
+    public boolean shouldReport(Set<String> metricsToIgnore, Function<String, Boolean> testChannelCheck) {
```
lucasmo, line 51: It feels a little weird to have a metric know whether or not it should report itself. I am really only saying this because the signature of this function feels funny to me.

lucasmo (approved):
My comments are all pretty minor and can probably be ignored aside from the stuff with `testTime`. Great test coverage!

src/main/java/com/flightstats/hub/filter/MetricsRequestFilter.java
lkemmerer: Yeah, I'm gonna do a little testing on dev before I switch it back to actually sending metrics to datadog. I got spooked after doing dumb things on Friday.

src/main/java/com/flightstats/hub/util/RequestMetric.java
lkemmerer: It made it easier to test, but then I refactored the calling method to make it more testable. Totally makes sense to move it back to where it belongs now that testing's not a pain.

src/main/java/com/flightstats/hub/util/RequestMetric.java
lkemmerer: I think it's just Java 9+. The fact that the method will eventually exist makes me happy, though. This is a crap way to do a simple thing.

src/main/java/com/flightstats/hub/util/HubRequest.java
lkemmerer: Hm, lost my comment here.

The builder and `HubRequest` proper got more complicated and weird when I realized Java says using `Optional<>` as attributes on a class is an antipattern (because it's a functional concept). While I understand that, I think perhaps doing it will make this code less confusing. I'll give it a try.

src/main/java/com/flightstats/hub/filter/MetricsRequestFilter.java
lucasmo: I guess you said that in the PR comment, which I clearly did not read.

src/main/java/com/flightstats/hub/metrics/StatsDFilter.java
```diff
@@ -67,8 +69,10 @@ public boolean isTestChannel(String channel) {
         return channel.toLowerCase().startsWith("test_");
     }
 
-    public Set<String> getRequestMetricsToIgnore() {
-        return requestMetricsToIgnore;
+    public boolean isIgnoredRequestMetric(RequestMetric metric) {
+        return metric.getMetricName()
+                .map(requestMetricsToIgnore::contains)
+                .orElse(true);
```
lucasmo, line 75: I think you mean `false` here?

lucasmo: Nevermind, this is saying "if there's no metric name, ignore it"

lkemmerer: I had `false` here initially but one of my stupid tests failed in protest. 😂

src/main/java/com/flightstats/hub/metrics/StatsdReporter.java
lkemmerer: done.

PR #1303, opened 2022-09-14
HUB-1175: Remove dynamic k8s part of system tests

lkemmerer:
We've removed the need to run our system tests as new releases in their own namespaces (we're using a static callback server and running tests against a full development deployment), so this rips out the Helm portion of our system tests. We can revisit  our strategy for doing so at a future point if ever there's a need to do so.

(this also does some log cleanup because the Jenkins output was _noisy_)


Paul-Hess (approved):

Paul-Hess:
LGTM

PR #1306, opened 2022-10-24
Remove some metrics that we probably don't need

lkemmerer:
This deletes:

`hub.get` -- which logs the time taken any time someone requests a single object using a full object key
`s3.list` -- metrics around how long and often we do lists against S3. This might be useful, but out of all the s3 actions, probably the least
`heisenberg` -- best I can tell is that it's the amount of time it takes for other nodes to respond that they've successfully stored an item in their spoke
`consistent` -- related to `heisenberg`, it's the time it takes for an item to have propagated to a quorum of hub node spokes

the only one of these that I ever look at is `s3.list`, but I don't think it's ever helped me solve a production issue...? If folks feel like it'd be good to keep around, I can remove that change.

### DIFF
--- a/src/main/java/com/flightstats/hub/spoke/SpokeManager.java
+++ b/src/main/java/com/flightstats/hub/spoke/SpokeManager.java
@@ -161,9 +161,6 @@ public class SpokeManager implements SpokeClusterHealthCheck, SpokeChronologySto
                         response = write_client.resource(uri).put(ClientResponse.class, payload);
                         traces.add(server, response.getEntity(String.class));
                         if (response.getStatus() == 201) {
-                            if (firstComplete.compareAndSet(false, true)) {
-                                statsdReporter.time(channel, "heisenberg", traces.getStart());
-                            }
                             quorumLatch.countDown();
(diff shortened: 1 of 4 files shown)


src/main/java/com/flightstats/hub/spoke/SpokeManager.java
```diff
@@ -161,9 +161,6 @@ public void run() {
                         response = write_client.resource(uri).put(ClientResponse.class, payload);
                         traces.add(server, response.getEntity(String.class));
                         if (response.getStatus() == 201) {
-                            if (firstComplete.compareAndSet(false, true)) {
```
lucasmo, line 164: Can we kill line 151 as well where we initialize `firstComplete`?

lucasmo (approved):
Looks pretty straightforward 👍

src/main/java/com/flightstats/hub/spoke/SpokeManager.java
lkemmerer: Might be a few days til I get around to this, but I'll put it/merging/deploying this on my to do list.

PR #1310, opened 2022-11-10
HUB-1203: Upgrade Zookeeper and test libraries

lkemmerer:
_(Please pardon the messiness of the commits. I didn't think to tidy them up prior to creating the PR. I think the changes are simple enough that that might be ok.)_

The story of this ticket goes something like:
1. Realize we're running ZK 3.4.11 rather than 3.5.9 (which is what we were running for the server).
2. Upgrade to 3.4.14 prior to bumping up to 3.5. This worked fine, though we (I) was doing some weird things with junit, so I updated and fixed those.
3. Update to 3.5.9. This broke some tests and it seemed to be due to the curator libraries we're using, where we were ignoring their ZK depenency.
4. Throw caution to the wind and update ZK all the way to 3.8 and the curator libraries up to their most recent versions, cutting out the special dependency management that was there. This, surprisingly, seems to be working with the 3.5.9 version of the ZK server.

Next steps before merging this in would be to update the server up to 3.8 on dev and see what happens. If there aren't any more code changes, I'll merge this and deploy it to staging. If there are necessary code changes, we'll need to downgrade the server back to 3.5.9 to ensure the changes are compatible. That way we can decouple code deploy from the server upgrade.

I've attached the gradle dependency diffs between master and this branch and then (maybe unhelpfully) added the more specific ZK/testing library diffs for [main/systemTests/tests.

[main-diff.txt](https://github.com/flightstats/hub/files/9983933/main-diff.txt)
[systemTest-diff.txt](https://github.com/flightstats/hub/files/9983934/systemTest-diff.txt)
[test-diff.txt](https://github.com/flightstats/hub/files/9983936/test-diff.txt)
[zk-dep-diff.txt](https://github.com/flightstats/hub/files/9983937/zk-dep-diff.txt)


build.gradle
```diff
@@ -125,17 +125,9 @@ dependencies {
...
-        exclude group: 'org.apache.zookeeper', module: 'zookeeper'
-    }
-    compile('org.apache.zookeeper:zookeeper:3.4.11')
-    //todo curator-test 4.0.0 has a compile issue with QuorumPeerConfig
-    compile('org.apache.curator:curator-test:2.12.0') {
-        exclude group: 'org.apache.zookeeper', module: 'zookeeper'
-    }
+    compile('org.apache.curator:curator-recipes:5.3.0')
+    compile('org.apache.zookeeper:zookeeper:3.8.0')
+    compile('org.apache.curator:curator-test:5.3.0')
```
lucasmo, line 130: can this be `testCompile`?

lucasmo (approved):
I can't say I looked at any of the test things in detail, but the gradle stuff makes sense 👍

build.gradle
lkemmerer: I'll take a look.

PR #1325, opened 2023-04-06
update ttl enforcer class with more efficient shell command, add a test

Paul-Hess:


src/main/java/com/flightstats/hub/dao/TtlEnforcer.java
```diff
@@ -37,9 +37,10 @@ public void deleteFilteredPaths(String path, ChannelService channelService,
                     .map(String::toLowerCase)
                     .filter(channel -> !channelSet.contains(channel.toLowerCase()) && !channel.equals(LOST_AND_FOUND_DIR))
                     .forEach(dir -> {
-                        String dirPath = path + "/" + dir;
-                        log.info("removing dir without channel {}", dirPath);
-                        commander.runInBash("rm " + " -rf " + dirPath, 1);
+
+                        String command = String.format("find %s/%s -type f --mmin +360 -delete", path, dir);
```
lucasmo, line 41: A few concerns:

- The path is not quoted, which is scary, but I guess it wasn't quoted before so this isn't a regression
- The previous rm -rf would also delete the directory, but this won't. I'm not sure if that matters.
- The previous rm -rf would delete all files; this only deletes files modified more than 6 hours ago.

It sounds like this change would:
- keep more files around
- not delete the directory

Is that intended?

Paul-Hess: I used the script essentially as it is running in chron currently as a patch.
I think the older than minute marker was a failsafe only for the patch version and should probably be removed.
I think adding in an unlink or rm -rf on the directory itself after the  files are deleted would be good and would not run into the same resource issue.

lucasmo: The one in cron looks like this:
```
# clean up read spoke for items outside the spoke window.
25 * * * * find /spoke/read -type f -mmin +360 -delete
```
which is for the entire spoke read cache. It looks like this change above would be doing that find individually for each directory. But I don't see how removing the old stuff with find becomes more efficient than removing it with rm -rf if we're going for an entire directory.

Is the case that this rm -rf is only ever called for spoke write cache, and not spoke read cache?

Maybe @lkemmerer can chime in as they added this to cron

Paul-Hess: talked with Laurie, closing this for now because I replaced this command in the wrong spot

Paul-Hess: Reopened. I think the changes should address the concerns above, lemme know.

lkemmerer:
The comment that Lucas made about the new command not cleaning up the empty directories *is* a concern that the quick fix we had done in the past (including the current prod cron job) doesn't address but we probably should. There are a bunch of empty read spoke `FinalizedFlightSnapshot` directories hanging out because of that. I don't know what the best way to clean that up would be, but it is something we need to make sure we do.

The other thing that we need to do is to modify the logic in `handleCleanup` in the Spoke TTL enforcer so that it uses the `else` path for the read spoke. Keeping it the way that it is now will continue to ignore anything that doesn't fall into the very specific timeframes that the `if` path looks at, which doesn't clean up most of historical data that gets stored in the read cache.

The read path is the main thing that needs optimizing and changing. The write cleanup works pretty well, if maybe a little slowly (maybe 5 minutes each run), but the TTL enforcer for the read cache was taking 15+ minutes to run and not actually getting all of the files cleaned up. ([See the Spoke TTL enforcer runtime for the past 3 months on iad.prod in Datadog for more info](https://app.datadoghq.com/dashboard/wsq-d9n-26w/hubs-by-cluster?fullscreen_end_ts=1680820683596&fullscreen_paused=false&fullscreen_section=overview&fullscreen_start_ts=1673044683596&fullscreen_widget=363156464&from_ts=1680806264451&to_ts=1680820664451&live=true))

lucasmo:
I believe `find <dir> -type d -empty` will find empty directories.

lucasmo:
```
~$ mkdir foo && cd foo
~/foo$ find . -type d -empty -depth +0
~/foo$ mkdir bar
~/foo$ find . -type d -empty -depth +0
./bar
~/foo$ mkdir bar/baz
~/foo$ find . -type d -empty -depth +0
./bar/baz
~/foo$ touch bar/baz/boo
~/foo$ find . -type d -empty -depth +0
~/foo$
```

Paul-Hess:
> ```
> ~$ mkdir foo && cd foo
> ~/foo$ find . -type d -empty -depth +0
> ~/foo$ mkdir bar
> ~/foo$ find . -type d -empty -depth +0
> ./bar
> ~/foo$ mkdir bar/baz
> ~/foo$ find . -type d -empty -depth +0
> ./bar/baz
> ~/foo$ touch bar/baz/boo
> ~/foo$ find . -type d -empty -depth +0
> ~/foo$
> ```

I assume you ran that on mac? 
I think the find on ubuntu is a little different I went with 
find $directory -depth -type d -empty -not -path "*lost+found*" -print -delete
I am not sure that -depth is required there because I believe it uses that behavior by default with the -delete flag

lucasmo:
> I assume you ran that on mac? I think the find on ubuntu is a little different I went with find $directory -depth -type d -empty -not -path "_lost+found_" -print -delete I am not sure that -depth is required there because I believe it uses that behavior by default with the -delete flag

Yeah I ran it on mac. If you don't specify depth of >0 it will include `.` if the root directory is empty, which is probably undesired behavior.

Paul-Hess:
> > I assume you ran that on mac? I think the find on ubuntu is a little different I went with find $directory -depth -type d -empty -not -path "_lost+found_" -print -delete I am not sure that -depth is required there because I believe it uses that behavior by default with the -delete flag
> 
> Yeah I ran it on mac. If you don't specify depth of >0 it will include `.` if the root directory is empty, which is probably undesired behavior.

ah yes, added mindepth 1 flag which is the equivalent on ubuntu

lkemmerer (approved):
I believe this resolves all the comments I had on the PR. I am mildly sad that there are no tests on the `SpokeTtlEnforcer`, but I understand that you'd be starting from scratch there if you wanted to exercise the new path and add tests around the old and unit tests are hard for the Hub...

PR #1329, opened 2023-05-11
Build tests

Paul-Hess:


src/test/java/com/flightstats/hub/webhook/WebhookLeaderLocksTest.java
```diff
@@ -46,20 +44,14 @@ void deletePath() throws Exception {
...
     void createWebhookLeader() throws Exception {
+        curator = IntegrationTestSetup.run().getZookeeperClient();
+        zooKeeperUtils = new SafeZooKeeperUtils(curator);
         try {
             createPath();
         } catch (NodeExistsException e) {
-            log.error(e.getMessage());
-            deletePath();
-            createPath();
+            log.error("webhook leader path already exists");
```
lkemmerer, line 54: Deleting and recreating the path was causing issues? I would have thought that clearing out the state would help with state bleed-over.

lkemmerer (approved):
I see the build passes. Woo!

src/test/java/com/flightstats/hub/webhook/WebhookLeaderLocksTest.java
```diff
@@ -4,10 +4,7 @@
 import com.flightstats.hub.util.SafeZooKeeperUtils;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.curator.framework.CuratorFramework;
-import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.*;
```
lkemmerer, line 7: Minor nit: I think we had at one point decided never to collapse imports by namespace. It's the single customization I have to remember to make in Intellj.

Paul-Hess: I have re-installed intellij and forgot to go in and fix that setting, thanks for catching that

src/test/java/com/flightstats/hub/webhook/WebhookLeaderLocksTest.java
Paul-Hess: It was not, no, it is not necessary though, moving the curator object creation into the before each instead of the before all isolated the state and allowed the delete path in the after each to work as intended

PR #1332, opened 2023-05-23
Dr client region

Paul-Hess:


src/main/java/com/flightstats/hub/config/properties/AwsProperties.java
```diff
@@ -35,4 +35,6 @@ public String getSigningRegion() {
         return propertiesLoader.getProperty("aws.signing_region", "us-east-1");
     }
 
+    public String getDisasterRecoveryRegion() { return propertiesLoader.getProperty("aws.disaster.recovery.region", "us-west-2"); }
```
lkemmerer, line 38: Just verifying that this does the right thing if there's no DR bucket. Looks like it does, since you're checking for bucket name in a couple places. I think staging/dev/integration tests all lack DR buckets?

lkemmerer (approved):
I guess we can't really test whether this works til we get it in an environment with a DR bucket. Things run ok in dev, though?

src/main/java/com/flightstats/hub/config/properties/AwsProperties.java
Paul-Hess: yeah, even if it gets called when there is no corollary bucket it will just return the default

PR #1334, opened 2023-06-12
Fix region

Paul-Hess:


lkemmerer:
I might've missed this in standup...apologies!...what was going on with the region? Was the DR manager still using the primary region?

Paul-Hess:
> I might've missed this in standup...apologies!...what was going on with the region? Was the DR manager still using the primary region?

On the first fix the DR manager was changed to use the correct region, the second fix was to isolate the singletons because importing two singletons of the same class was causing one to overwrite the other, the effect that I saw was the main bucket having the incorrect region but it was probably a race condition and either one could have had the incorrect region depending on whatever factors at startup.

lkemmerer (approved):

PR #1337, opened 2023-07-18
Try a new idea: Eagerly initialize a singleton

lucasmo:
This should own both maintenance managers.


lucasmo:
I have validated this works in dev.d

PR #1343, opened 2024-02-14
added condition to ignore ttl days and maxItems if KeepForever is tru…

harinadh-dasari:
…e , in case of false throw Exception


lkemmerer (approved):
Looks good! Thanks for adding the extra test cases around keepForever!

PrabhavathyG (approved):

PrabhavathyG (approved):

PR #1351, opened 2024-03-06
HUB-1339: Initial Fix to check for InternalSpokeResource

mishraa8:
This fix is for testing whole flow for Checkmarx.
Also to fix one of Medium cve.


lkemmerer (approved):
Looks good. As a reminder, don't forget to at least run the build-hub Jenkins job (and maybe deploy and run cluster integration tests on dev) prior to merging this. (Ideally that job would be triggered on all pushes and a passing build would be a gate to merging, but until it is...)

mishraa8:
yes, it should be. here is build 
https://ddt-jenkins.prod.flightstats.io/view/hub%20build%20and%20deploy/job/run-hub-development-integration/71/

harinadh-dasari (approved):
Looks Good

PR #1352, opened 2024-03-07
HUB-1287 Added condition to alter the spoke key path to support windo…

PrabhavathyG:


lkemmerer:
Do we have channel validation such that it's not legal to put slashes in the name? I believe so, but wanted to double check.

src/test/java/com/flightstats/hub/spoke/FileSpokeStoreTest.java
```diff
@@ -47,12 +47,12 @@ void testWriteRead() {
 
     @Test
     void testPathTranslation() {
-        String incoming = "/test_0_4274725520517677/2014/11/18/00/57/24/015/NV2cl5";
+        String incoming = "test_0_4274725520517677/2014/11/18/00/57/24/015/NV2cl5";
```
lkemmerer, line 50: Does this test still pass on the Jenkins server without the leading slash?

PrabhavathyG:
Approach changed. Hence closing this PR.

PR #1355, opened 2024-03-11
Feature/hub 1333

PrabhavathyG:
- Added FileSystems.getDefault().getSeparator() method to replace the harcoded unix based "/" file separator.
- Modified url strings to use Paths.get()

### DIFF
--- a/src/main/java/com/flightstats/hub/config/properties/SpokeProperties.java
+++ b/src/main/java/com/flightstats/hub/config/properties/SpokeProperties.java
@@ -4,6 +4,8 @@ import com.flightstats.hub.spoke.SpokeStore;
 import org.apache.commons.lang3.StringUtils;
 
 import javax.inject.Inject;
+import java.nio.file.FileSystems;
+import java.nio.file.Paths;
 
 public class SpokeProperties {
 
@@ -27,8 +29,8 @@ public class SpokeProperties {
     public String getPath(SpokeStore spokeStore) {
         String property = "spoke." + spokeStore + ".path";
         String fallbackProperty = "spoke.path";
-        String defaultPath = "/spoke/" + spokeStore;
-        return propertiesLoader.getProperty(property, propertiesLoader.getProperty(fallbackProperty, defaultPath));
+        String defaultPath = Paths.get(FileSystems.getDefault().getSeparator(),"spoke", String.valueOf(spokeStore)).toString();
+        return Paths.get(propertiesLoader.getProperty(property, propertiesLoader.getProperty(fallbackProperty, defaultPath))).toString();
     }
 
     public String getStoragePath() {
--- a/src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
+++ b/src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
@@ -253,7 +253,7 @@ public class FileSpokeStore {
                 String spokeKeyFromPath = spokeKeyFromPath(Paths.get(hoursPath,minute,fileName).toString());
                 log.trace("looking at file {} ", spokeKeyFromPath);
                 if (Paths.get(spokeKeyFromPath).toString().compareTo(Paths.get(limitPath).toString()) < 0) {
-                    return channel + "/" + spokeKeyFromPath;
+                    return Paths.get(channel,spokeKeyFromPath).toString();
                 }
             }
         }
--- a/src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
+++ b/src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
@@ -2,6 +2,7 @@ package com.flightstats.hub.spoke;
 
 import com.flightstats.hub.model.ContentKey;
 import com.flightstats.hub.model.MinutePath;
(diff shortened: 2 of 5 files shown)


src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
```diff
@@ -244,9 +250,9 @@ private String getLatest(String channel, String limitPath, DateTime hourToSearch
                     .collect(Collectors.toList());
             Collections.reverse(fileNames);
             for (String fileName: fileNames) {
-                String spokeKeyFromPath = spokeKeyFromPath(hoursPath + "/" + minute + "/" + fileName);
+                String spokeKeyFromPath = spokeKeyFromPath(Paths.get(hoursPath,minute,fileName).toString());
                 log.trace("looking at file {} ", spokeKeyFromPath);
-                if (spokeKeyFromPath.compareTo(limitPath) < 0) {
+                if (Paths.get(spokeKeyFromPath).toString().compareTo(Paths.get(limitPath).toString()) < 0) {
                     return channel + "/" + spokeKeyFromPath;
```
mishraa8, line 256: This also needs to modified as Paths.get(channel, spokeKeyFromPath).toString

src/test/java/com/flightstats/hub/spoke/FileSpokeStoreTest.java
```diff
@@ -47,14 +49,14 @@ void testWriteRead() {
 
     @Test
     void testPathTranslation() {
-        String incoming = "/test_0_4274725520517677/2014/11/18/00/57/24/015/NV2cl5";
+        String incoming = "test_0_4274725520517677/2014/11/18/00/57/24/015/NV2cl5";
         File outputFile = spokeStore.spokeFilePathPart(incoming);
         String filePath = "/test_0_4274725520517677/2014/11/18/00/57/24015NV2cl5";
```
mishraa8, line 54: Shouldn't this also require modification as per Paths?

lkemmerer (approved):
My suggestions are optional. I leave it up to you to determine if the added abstraction is worth it.

Please be aware that this and Abhay's PR have some overlap, so you'll need to collaborate to make sure what's merged to develop doesn't contain two ways of doing the same thing.

src/main/java/com/flightstats/hub/spoke/FileSpokeStore.java
```diff
@@ -161,21 +165,23 @@ File spokeFilePathPart(String urlPathPart) {
...
+        int i = path.lastIndexOf(fileSeparator);
         String suffix = path.substring(i + 1);
         if (suffix.length() > 4) {
             // presence of second proves file aims at a full payload path
             String folderPath = path.substring(0, i);
             String seconds = suffix.substring(0, 2);
             String milliseconds = suffix.substring(2, 5);
             String hash = suffix.substring(5);
-            return folderPath + "/" + seconds + "/" + milliseconds + "/" + hash;
+            return Paths.get(folderPath ,seconds , milliseconds , hash).toString();
```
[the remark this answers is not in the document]
lkemmerer: whoops, I guess we're not normalizing anything here, so perhaps a different name would be in order here.

src/main/java/com/flightstats/hub/config/properties/SpokeProperties.java
```diff
@@ -27,8 +29,8 @@ public int getTtlMinutes(SpokeStore spokeStore) {
     public String getPath(SpokeStore spokeStore) {
         String property = "spoke." + spokeStore + ".path";
         String fallbackProperty = "spoke.path";
-        String defaultPath = "/spoke/" + spokeStore;
-        return propertiesLoader.getProperty(property, propertiesLoader.getProperty(fallbackProperty, defaultPath));
+        String defaultPath = Paths.get(FileSystems.getDefault().getSeparator(),"spoke", String.valueOf(spokeStore)).toString();
```
ashoklxnx, line 32: "spoke" - Instead of hardcoding string here, can this also be accessed using a variable from  SpokePathUtil class ?

ashoklxnx (commented):
Added inline comment

src/main/java/com/flightstats/hub/config/properties/SpokeProperties.java
PrabhavathyG: Since this file belongs to load the spoke properties value from ***hub.properties file or default value if not present , we are mentioning(hardcoding) all the keys over here. But the corresponding values are picked from the properties file. So am leaving this as it is.

mishraa8 (approved):
Approved

src/main/java/com/flightstats/hub/util/HubUtils.java
```diff
@@ -43,6 +45,7 @@ public class HubUtils {
     private final Cluster hubCluster;
     private final ObjectMapper objectMapper;
     private final String uriScheme;
+    public static final String FILE_SYSTEM_SEPARATOR = FileSystems.getDefault().getSeparator();
```
PrabhavathyG, line 48: I am adding this constant in HubUtils as SpokePathUtils is not public class , due to that could not access outside package.

PR #1356, opened 2024-03-13
Releasing HUB-1333 and HUB-1347

mishraa8:


lkemmerer:
Is the commit history going to get weirder and weirder because we didn't do a fast-forward merge of develop into master last time? I'd like us to figure out how to get things into a state that master's history is the same as develop up to new work that we want to merge in. Maybe to get there we would need to do a one-time reset of `develop` to what's on `master`? I'm not sure.

lkemmerer:
To help this from happening in the future, I updated the master branch to require linear history -- https://github.com/flightstats/hub/settings/branch_protection_rules/2131062 I guess we can see how merging this into master goes based on that restriction?

src/main/java/com/flightstats/hub/channel/ChannelResource.java
```diff
@@ -214,18 +214,6 @@ public Response insertValue(@PathParam("channel") String channelName,
         }
     }
 
-    @POST
-    @Consumes("multipart/*")
-    @Produces(MediaType.APPLICATION_JSON)
-    @Path("/batch")
-    public Response insertBatch(@PathParam("channel") final String channelName,
```
lkemmerer, line 221: These changes should already be in master, no?

mishraa8:
Yes, Agree with your point @lkemmerer , master and develop branch history would never be same, and it can go worse.
We can do deleting Develop branch and create Develop branch again. Hope that will make it clear. Going forward with Linear History enabled it will be clean and same as master history.

PR #1365, opened 2024-03-19
Added Encoder for Channelname in delete method

Haripriya99j:
Checkmarx High Vulnerabilities Fix:
Added Encoder for channelName user_input in delete method.
InternalChannelResource.java

### DIFF
--- a/src/main/java/com/flightstats/hub/channel/InternalChannelResource.java
+++ b/src/main/java/com/flightstats/hub/channel/InternalChannelResource.java
@@ -14,6 +14,7 @@ import com.flightstats.hub.util.StaleEntity;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.joda.time.DateTime;
+import org.owasp.encoder.Encode;
 
 import javax.inject.Inject;
 import javax.ws.rs.DELETE;
@@ -192,7 +193,7 @@ public class InternalChannelResource {
         if (channelService.delete(channelName)) {
             return Response.status(Response.Status.ACCEPTED).build();
         } else {
-            return Response.status(Response.Status.NOT_FOUND).entity("channel " + channelName + " not found").build();
+            return Response.status(Response.Status.NOT_FOUND).entity("channel " + Encode.forHtml(channelName) + " not found").build();
         }
     }
 }
--- a/src/main/java/com/flightstats/hub/channel/InternalChannelResource.java
+++ b/src/main/java/com/flightstats/hub/channel/InternalChannelResource.java
@@ -116,20 +116,18 @@ public class InternalChannelResource {
     @Path("{channel}")
     @DELETE
     public Response delete(@PathParam("channel") final String channelName) {
-        //added sanitizing method for user input for channelname to prevent XSS Vulnerability
-        String sanitizedChannelName = Encode.forUriComponent(channelName);
-        channelService.getChannelConfig(sanitizedChannelName, false)
+        channelService.getChannelConfig(channelName, false)
                 .orElseThrow(() -> {
-                    Response errorResponse = ChannelResource.notFound(sanitizedChannelName);
+                    Response errorResponse = ChannelResource.notFound(channelName);
                     throw new WebApplicationException(errorResponse);
                 });
 
         if (contentProperties.isChannelProtectionEnabled()) {
-            log.info("using internal localhost only to delete {}", sanitizedChannelName);
-            return LocalHostOnly.getResponse(uriInfo, () -> deleteChannel(sanitizedChannelName));
+            log.info("using internal localhost only to delete {}", channelName);
+            return LocalHostOnly.getResponse(uriInfo, () -> deleteChannel(channelName));
         }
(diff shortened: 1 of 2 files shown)


src/main/java/com/flightstats/hub/channel/InternalChannelResource.java
```diff
@@ -114,18 +116,20 @@ public Response refresh(@QueryParam("all") @DefaultValue("true") boolean all) {
...
-                    Response errorResponse = ChannelResource.notFound(channelName);
+                    Response errorResponse = ChannelResource.notFound(sanitizedChannelName);
                     throw new WebApplicationException(errorResponse);
                 });
 
         if (contentProperties.isChannelProtectionEnabled()) {
-            log.info("using internal localhost only to delete {}", channelName);
-            return LocalHostOnly.getResponse(uriInfo, () -> deleteChannel(channelName));
+            log.info("using internal localhost only to delete {}", sanitizedChannelName);
+            return LocalHostOnly.getResponse(uriInfo, () -> deleteChannel(sanitizedChannelName));
```
PrabhavathyG, line 129: Should we need sanitizedChannelName in all instances ? deleteChannel() method uses channelName , which should not be the encoded one I guess.

lkemmerer: Sanitized and normal should be the same for any valid channel names, I think. There's a validation in ChannelValidator that uses the following regex for name: `public static final String VALID_NAME = "^[a-zA-Z0-9_-]+$";` (in `ContentConstant.java`). So...I think using either one should be ok in this case.

lkemmerer (commented):
As a best practice, until there's a way to automate build checks, I think it'd be helpful to include a link to a successful `build-hub` job in Jenkins. Taking a a look there, the latest build for this branch failed: https://ddt-jenkins.prod.flightstats.io/job/build-hub/406/

src/main/java/com/flightstats/hub/channel/InternalChannelResource.java
mishraa8: yes, Sanitized and Normal should work same for any valid Channel Names. This is only applied to Public method which are exposing APIs. 
This can be tried. That method is private and can be moved into HubUtils and update references accordingly.
The only worry I have here, is while putting Channel Name in Spoke/S3, it has be encoded in same way.

lkemmerer: I'm not sure I fully understand this -- I don't think we need to use the validation I mentioned anywhere new, since it's being used on creation and update of channel configs. I was just pointing out that I believe `A-Z` + `_` and `-` are all characters that would not be impacted by url encoding, so I think encoded would always equal unencoded at the point that we're interacting with S3 or spoke. I might be missing something?

mishraa8: This Channel Validator is using above regex to validate the channel name against A-Z + _ and -. And Encode.forUriComponent - is only doing against ChannelName. So yes agree, it is safe to use. It wont break anything. Now I got whole point here mentioning about it in your first comment. 

@Haripriya99j  - we can go with this change, just remove StringEscapeUtils.escapeHtml4, which is not needed. Rest all  good.

Haripriya99j: Yes Abhay, Even I also approached in this way only. Once it goes well, I will raise another PR request with above mentioned changes.

Haripriya99j: Closing this PR

Haripriya99j: Using "StringEscapeUtils.escapeHtml4" only fix can be resolved. I just tried with many combination of encode snippets but still this only resolving the issue.
And There's a validation in ChannelValidator that uses the following regex for name: public static final String VALID_NAME = "^[a-zA-Z0-9_-]+$"; (in ContentConstant.java).
Since the issue is in InternalChannelResource.java, we need to add separate validator there also so that only it can be fixed for vulnerabilities. And here suggesting to go with "StringEscapeUtils.escapeHtml4".

lkemmerer:
Other than adding the owasp library to the build and linking to a build with a passing build and integration tests, this looks good!

mishraa8 (approved):
Can you attach successful build job id and checkmarx screen shot.

PrabhavathyG (approved):

PrabhavathyG (approved):

mishraa8:
Good work @Haripriya99j!! You actually found actual issue causing this vulnerability, which resolved other vulnerabilities as well. 
Let's follow this other similar vulnerability, if any exist.

mishraa8 (approved):
Good Work.
Let me know when you want to merge.

Haripriya99j:
> Can you attach successful build job id and checkmarx screen shot.

https://ddt-jenkins.prod.flightstats.io/job/run-hub-development-integration/112/

PrabhavathyG (approved):

PR #1366, opened 2024-03-22
Added encodeURI function in toc.html and deleted docs/js/mydoc_scroll.html

Haripriya99j:
Added encodeURI function in toc.html and deleted docs/js/mydoc_scroll.html


mishraa8 (approved):
It is good, it is going with separate commit.
All looks good to me as per analysis.

lkemmerer (approved):

harinadh-dasari (approved):
LGTM

PR #1375, opened 2024-04-29
Hub 1358 check marx high fix

harinadh-dasari:


allustag (approved):

src/main/java/com/flightstats/hub/channel/TagContentResource.java
```diff
@@ -166,16 +167,18 @@ public Response getValue(@PathParam("tag") String tag,
                              @PathParam("ms") int millis,
                              @PathParam("hash") String hash,
                              @HeaderParam("Accept") String accept) {
-        ContentKey key = new ContentKey(year, month, day, hour, minute, second, millis, hash);
+        String encodedHtmlTag = Encode.forHtml(tag);
```
allustag, line 170: instead of introducing new variables here, can we try with below
tag = Encode.forHtml(tag);
hash = Encode.forHtml(hash);

